Issue #22, opened 2015-02-06
Which Pointing?


drphilmarshall:
We'll need to choose a pointing for any given Twinkles field we make. What criteria should we choose? CatSim objects in that field? OpSim observing pattern in that field? Product for this hack could a better markdown document, with links to resources. Or a small piece of code that visualized the properties of any given pointing, both in terms of CatSim and OpSIm outputs.

SimonKrughoff:
This is also a question of what run to use.  The easiest to use is OpSim 3.61, but it would be nice to make it configurable.

wmwv:
OpSim 3.61

One field in deep drilling survey from list of guaranteed files
http://www.lsst.org/News/enews/deep-drilling-201202.html

Extended Chandra Deep Field - South
(RA, Dec) = (03:32:30, 10:00:24)
(l,b) = (224.07, -54.47)

drphilmarshall:
Nice! Thanks Michael.

If we go with this, can we straightforwardly downsample to just the main
WDF proposal cadence as well? Are both proposals in play for the DDFs?

wmwv:
Yes, the deep-drilling fields are also covered independently by the Main (WDF) survey.  These an be extracted independently from OpSim.

SimonKrughoff:
There's a copy/paste error in the coordinates in this issue.  It should be:
RA: 03h 32m 30s Dec: -28d 06m 00s

I reopened to remind us to fix this in the documentation, but we can open another issue if you want.

wmwv:
@SimonKrughoff   Thank you!  Apologies, I copy+pasted the RA of each of the Extended Field Chandra South Field and the COSMOS field, instead of the (RA, Dec) of the former.

SimonKrughoff:
I only found one location this was mentioned and it really is where we pointed, so I didn't change it.  I did add the pointing to the survey design document.  Assuming PR #48 is approved I'm closing this issue.

PR #48, opened 2015-10-28
Add location of the field to the survey design

SimonKrughoff:
In issue #22 the Extended Chandra Deep Field South was chosen as the best option.


SimonKrughoff:
This seems very non-controversial, so I will merge this by the end of lunch if no one disagrees.

Issue #32, opened 2015-10-22
Consistent seeding for sprinkled objects at different pointings in time


jbkalmbach:
This is the problem @danielsf brought up in #31 concerning lenses.

I'm moving it here so that we can separate it from that pull request since this is a problem that is also being worked on by @rbiswas4 for supernovae.

jbkalmbach:
Added to task list under #37

jbkalmbach:
After discussion with @rbiswas4 and @danielsf we decided that since the number of galaxy objects in the proposed first run of Twinkles field is on the order of 250,000 we are safe using the result of (galid % (2^32 -1)) as the seed for numpy's random number generator. Rahul has also shown that none of these values repeats within the field of interest.

rbiswas4:
As mentioned by @jbkalmbach,  http://nbviewer.ipython.org/gist/rbiswas4/b1952341e75a049f820f shows that in the patch of sky used by @SimonKrughoff contains a small number of galaxies, and the galtileid % (2^32 -1) values are unique

jbkalmbach:
#102 incorporates this into the sprinkler.

PR #102, opened 2016-01-30
Consistent Seeding Added

jbkalmbach:
Random number seeding is now consistent so the same lenses get sprinkled every visit. Also, finally got database timeout right.

This incorporates the random numbers as outlined in #32


drphilmarshall:
Nice! Glad you nailed this one :-)

PR #121, opened 2016-02-12
Added materials discussing strategy of selection of  

rbiswas4:
Added a file showing Kraken visits of the Twinkles region and a proposal on which obshistIDs that we should attempt. At UW, we would like to generate PhoSim Instance Catalogs for these visits this weekend. So, if someone notices a problem with this proposal, or would like to suggest something else please let us know quickly.


drphilmarshall:
Interesting! Love the [notebook](https://github.com/DarkEnergyScienceCollaboration/Twinkles/blob/d68df56be68828d98a13dfeba7341b399054a157/code/SN_sims/Kraken_vistsSelection.ipynb)  From your plots I make it that season 2 runs from about MJD60190 to MJD60330, and so lasts about 140 days, or about 4.6 months. Good! In that period there are 45 nights with at least one visit, and it looks like about 21 of them have all 5 of the *grizy* filters, while the other 24 nights are spent hammering away at the u-band. If we just keep one visit per filter per night, the total number of visit images looks like being about 21x5 + 24 = 129 or so. This is pretty small compared to Tom's 1000 per week capacity: but maybe still big enough to get some nice light curves and analyze them? @wmwv ?

The night-to-night cadence in this DDF looks like being 140/(21+24) = 3 days, so I guess yes, these light curves will be nice! (Just not very deep.) Thanks you two!

PR #123, opened 2016-02-16
Take custom list of visit IDs

jbkalmbach:


drphilmarshall:
Excellent! You must be getting close to making some instance catalogs up
there :-)

Issue #105, opened 2016-02-01
repository reorganization


jchiang87:
It's probably worth re-organizing this repository in order to provide an example for other DESC repos.  
One thing to consider is making a python folder and putting any code that is mean to be imported from a `twinkles` module to live as a subfolder.    This will make it easier/possible for whatever package management tool we use, e.g., eups, to know what to add to the user's PYTHONPATH when setting up the package.  The same would apply to the new [Monitor](https://github.com/DarkEnergyScienceCollaboration/Monitor) and [SLTimer](https://github.com/DarkEnergyScienceCollaboration/SLTimer) repos.  
@SimonKrughoff You mentioned that you were going to look into reorganizing this repo.  I'm wondering if we should just make it eups-compliant.

SimonKrughoff:
@jchiang87 thanks for the nudge.  I have been thinking a little about this.  I have a task on my plate to re-organize the repo and was wondering whether we want to copy DM's namespace conventions.  I.e. twinkles goes to `lsst.twinkles` or `desc.twinkles`.  Either way, I think having a python directory with the code inside that would be a big win if we can standardize on it.

jchiang87:
I think that `desc.twinkles` would be a good convention to adopt, but we should probably not going any deeper than that to avoid the tendency towards overly large packages with lots of subcomponents, e.g., afw.

wmwv:
I agree that `desc.` should be the base.

drphilmarshall:
Me too. Thanks Jim.

On Tue, Feb 2, 2016 at 6:49 AM, Michael Wood-Vasey <notifications@github.com

drphilmarshall:
Thoughts from discussion between @rbiswas4 and @jchiang87 : provide setup script to set environment variables. Implement importable python structure as discussed. Do this in a branch so we can read the new README and view the proposed structure?

jchiang87:
Closed by #147

PR #147, opened 2016-02-26
Reorganize

jchiang87:
Hi All,

This branch isn't ready yet to merge.  I just wanted to open the PR so that discussion on these changes could happen here instead of at #105.

This afternoon @rbiswas4 and I went through the folders and (somewhat aggressively) moved things to what may be more appropriate folders, assuming a LSST Stack/eups-based organization:  
- Files that looked like importable python modules were put in `python/desc/twinkles`.
- Files that looked like standalone scripts were put in `bin`.
- We created an `examples` folder and moved the python notebooks there.
- The cookbook and other documentation were moved to the `doc` folder.
- We added a `data` directory for `.csv` and a FITS files, all of which may be better suited to git-lfs.
- An `ups` folder with the associated eups files were added.
- A straw-man `setup/setup.sh` script was added as a first cut at getting things going with the Stack.  It will actually work if one already has the `OM10` package in the `PYTHONPATH` and a default version of `lsst_sims` has been set up.  Some details still need to be sorted out.
- Some files were `git rm`'d.
- The `workflows` folder  was left untouched except that we moved the physics override file to the `workflows/Twinkles-phoSim` folder.

Have a look at the [reorganize branch](https://github.com/DarkEnergyScienceCollaboration/Twinkles/tree/reorganize), and post comments here or on the `Files changed` tab.


wmwv:
The current template for LSST Stack repos is

http://github.com/lsst/templates

You may find it helpful to look through this, or perhaps you already have.

jchiang87:
Hi Michael!

I actually wasn't aware of that repo, so thanks indeed for the pointer.  I've been working on a cookiecutter project for building C++ code against the stack for use by DESC developers:

https://github.com/jchiang87/stack-build-template

If you have time to have a look, I'd welcome any comments or suggestions.

drphilmarshall:
I made some  further enhancements, to be considered as suggestions but just pushed so you can see them:

* README files in all folders
* Links fixed from top level README
* Better credits in same
* Renamed `Twinkles-phoSim` to `PhoSim` in workflow folder, for consistency (and the name Twinkles is already in the path)
* Contacts added where appropriate
* Invitation to issue us if the examples don't work...

jbkalmbach:
Hi all,
So I just took a look at this. Do we want to keep one or the other of InstcatGenerator.py or generatePhosimInput.py in the python/desc/twinkles folder? It seems like they are just two different ways of creating the sprinkled catalogs. Although I don't think InstcatGenerator has SNe.

jchiang87:
I was willing to remove InstcatGenerator, but @rbiswas4 was thinking that it should be adopted in generatePhosimInput.py after updating to the current set of sources that are available.  That plan made sense to me.   I was thinking that generatePhosimInput.py would then go into the `bin` folder.  For now, I think we can just keep both and unify them after merging this branch.

Any other issues to address before doing this merge?

jbkalmbach:
I'm not sure this need to be answered before merging but my only other question is what's the best way to connect the sprinkler/generators to the data files they need?

rbiswas4:
@jchiang87, yes, I think leaving it there will not hurt for now. We can decide about the best way forward when we have updated modules.

@jbkalmbach I think this depends on the size of an example dataset. If we can have a small example dataset (small such that it can be checked into the regular github repo without concern), we could use a here statement to access example data directories. This is useful to have working examples under version control, and tests. If such a small example directory cannot be built, (example for many tests we require too much data, fits files cannot be small enough, etc.), we might want to check things into git-lfs and I am not sure of the best way.

drphilmarshall:
@jchiang87 We'll need to merge *from* master into reorganize somehow, to capture all your postage stamp work from #149. Hopefully this will not be too painful...

jchiang87:
I'll have a look at this now.

jchiang87:
ok, @drphilmarshall, I think I got everything.

SimonKrughoff:
@jchiang87 this looks pretty good to me.  I see you've put all the files I had in there in reasonable spots, so I'd say merge away.

jchiang87:
@drphilmarshall I'll leave it to you to pull the trigger on this.  I _think_ we've gotten the nod from everyone who has written code so far.

drphilmarshall:
THE NEW ERA OF TWINKLES CODE ORGANIZATION HAS BEGUN

drphilmarshall:
Pull carefully, people - and feel free to ask for help if things seem to have gone weird. You might want to make a safe branch to try things out:
```
git checkout -b temp
git pull origin master
```
and then see what happens there. You can `git branch -D` to remove `temp` when you're set.

Issue #141, opened 2016-02-25
Create postage stamps


drphilmarshall:
Given DM `CoaddTempExp` images (that include image extensions for science, variance and mask, plus a common WCS header), extract a square cutout measuring X arcsec on a side centered on (Ra,Dec) as another `CoaddTempExp` object. In the first instance we will want to do this from a notebook just to make plots; later we may want to write these cutouts out to file and serve them from the data catalog? Not sure.

PR #149, opened 2016-02-26
Issue/141/create postage stamps

jchiang87:
#141:

Here is the first version.  We need to decide how to display things in a notebook.  The example code uses ds9, but presumably we'd want to use matplotlib.


drphilmarshall:
Let's use matplotlib since it is the prevailiing community standard - example code here:

https://github.com/drphilmarshall/StatisticalMethods/blob/master/examples/XrayImage/FirstLook.ipynb

drphilmarshall:
Notes from conversation:

* Let's start a `Display` module with a static function for using matpotlib to show FITS images. This can also be where we make light curve plots, eh @rbiswas4 ? And also color composite creation ( @SimonKrughoff )
* Enable array inputs (and hence array outputs): sometimes we'll want all stamps at one ra,dec from N exposures, and other times we'll want all stamps from one exposure at N ra,dec positions...
* Display methods will also need to handle array inputs (and outputs)

drphilmarshall:
Also:

* [ ] Check in small FITS image
* [ ] Check in tests

rbiswas4:
> Let's start a Display module with a static function for using matpotlib to show FITS images. This can also be where we make light curve plots, eh @rbiswas4 

Sounds good to me: Of course, this will import and heavily depend on @jbkalmbach's project where we are assembling the truth information.

jchiang87:
I added a`Display` module, test code, array generation, etc..  I think this is ready to consider merging into master.  Please have a look.

drphilmarshall:
This looks good to me, Jim! The notebook is great. I guess this will break the reorg pull request #147 - we'd better merge into the reorganize branch and consider merging that PR soon, too...

jchiang87:
@drphilmarshall Is that the usual way of handling these things?  I was thinking we could just merge this into master and then deal with conflicts over at #147 when it's time to merge that PR (which I agree we should do right away).  Otherwise the work in this branch may get held up in the reorganize branch if there are other issues to deal with there.

drphilmarshall:
I think you are saying exactly what I was trying to say, but better :-) Merging this PR (#149) now!

PR #164, opened 2016-03-04
Making Twinkles work in the Conda DM installation

drphilmarshall:
This seems important, if we are going to enable casual DM users like myself to contribute to Twinkles. This PR thread is to show you my attempts.

PR #165, opened 2016-03-05
added a few things to INSTALL.md

rbiswas4:
Added a few things for the SNObj code to work. I don't see a clear problem in using LSST_DIR as an environment but would tend to be wary of this, since that is the kind of variable that is automatically setup by eups, and so there might be unforseen clashes.


drphilmarshall:
This too has been superceded by work in #164 and elsewhere...

Issue #169, opened 2016-03-09
Connect Twinkles to Travis CI using Mario's conda install script for lsst-sims


jchiang87:
Mario describes this work over at [community.lsst.org](https://community.lsst.org/t/how-to-use-the-stack-with-travisci/587) and in his [github repo](https://github.com/mjuric/lsst-travis-demo).

We should also rewrite `python/desc/twinkles/__init__.py` so that it doesn't import everything by default to save some time for client code that doesn't need the catsim tools.

mjuric:
@jchiang87 et al: I'd be happy to consult & help!

jchiang87:
@mjuric Thanks for putting the travis-ci demo and conda distribution together! I think they will be a big help in getting the DESC repos working with travis-ci. 

I was able to get things working over in my fork (where I have admin privileges to hook up to travis-ci) and issued the PR (#170).  I  did need to install `pandas` (and `nose`) on top of the conda python, but other than that, everything else basically worked out of the box.

jchiang87:
Hey Phil!  Please hook us up to Travis CI.   You can close this once we get a travis passing badge on the readme! ;-)

drphilmarshall:
Done! [Check it out.](https://github.com/DarkEnergyScienceCollaboration/Twinkles/blob/master/README.md#latest-news)

PR #170, opened 2016-03-09
Hook up Twinkles to Travis CI using conda install of the stack + lsst_sims

jchiang87:
This addresses #169


.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: C
+
+install:
+  - CHANNEL=http://eupsforge.net/conda/dev ./setup/install.sh lsst-sims
```
mjuric: I'd propose to change this to:
```
CHANNEL=http://eupsforge.net/conda/dev ./setup/install.sh lsst-sims nose pandas
```
That way you don't need `requirements.txt`, and once pandas & nose are installed they'll be efficiently cached together with the rest of Python, etc.

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+  - export PATH="$HOME/miniconda/bin:$PATH"
+  - source eups-setups.sh
+  - setup lsst_sims
+  - pip install -r setup/requirements.txt
```
mjuric: If you implement the suggestion above, you can remove this line (and `requirements.txt`).

jchiang87: Awesome! Thanks!

.travis.yml
jchiang87: So I changed the line as suggested (86c0f15), and the build failed with this:
```
+conda config --add channels http://eupsforge.net/conda/dev
+conda install -q --yes 'lsst-sims nose pandas'
Error: Invalid package specification: lsst-sims nose pandas
The command "CHANNEL=http://eupsforge.net/conda/dev ./setup/install.sh lsst-sims nose pandas" failed and exited with 1 during .
Your build has been stopped.
```

jchiang87: Removing the double quotes around `$PACKAGE` at [line 59](https://github.com/jchiang87/Twinkles/blob/u/jchiang/connecting_to_travis-ci/setup/install.sh#L59) of `install.sh` fixed the problem.

mjuric:
@jchiang87 I fixed the mistake with the quoting on the current master (https://github.com/mjuric/lsst-travis-demo/blob/master/ci/install.sh), slightly differently than you did (I wanted to ensure any quoting or spaces in arguments survive and get passed on to `conda install`). It should be a drop-in replacement of the current `install.sh` file.

jchiang87:
Thanks @mjuric, I'll integrate your changes later today.  I'm wondering if it would be possible to "centralize" this script somehow so that updates are propagated to everyone's packages automatically, i.e., grab `install.sh` from your repo instead of using a local copy.

mjuric:
Yes, I've been thinking about how to do that, but I'm not sure how I'd do that (need to think more). Copying should be OK for now.

jchiang87:
Phil says ok, let's merge!

Issue #167, opened 2016-03-05
Photometric and Astrometric Validation of PhoSim runs


wmwv:
Compute and analyze overall astrometric and photometric repeatability of Twinkles PhoSim runs as part of Twinkles Run 1.

Tools will include `validate_drp`
https://github.com/lsst/validate_drp

and related tests, such as those @SimonKrughoff did in 
https://github.com/DarkEnergyScienceCollaboration/Twinkles/issues/70#issuecomment-168073498

This Issues is not directly looking at lightcurves of known variable objects.  Lightcurve analysis and tests may likely fall more appropriately under Issues for Monitor discussion.

https://github.com/DarkEnergyScienceCollaboration/Monitor

wmwv:
Here are some figures of the astrometric and photometric repeatability as compared to the LSST Science Requirements Document.

u: PA1(IQR) = 82.61+-1.53 mmag
i: PA1(IQR) = 21.95+-1.57 mmag
z: PA1(IQR) = 16.69+-0.57 mmag
y: PA1(IQR) = 25.54+-1.21 mmag

![twinkles_run1_output_i_am1_d_5_arcmin_17 0-21 5](https://cloud.githubusercontent.com/assets/3201257/13550091/7301be60-e2e3-11e5-930a-c9f9ba25259d.png)

![twinkles_run1_output_i_check_photometry](https://cloud.githubusercontent.com/assets/3201257/13550092/7d80f298-e2e3-11e5-8eaf-11db9279f6a1.png)
![twinkles_run1_output_i_pa1](https://cloud.githubusercontent.com/assets/3201257/13550094/8bbbb1c2-e2e3-11e5-89e9-1146f93f9b1e.png)

![twinkles_run1_output_i_check_astrometry](https://cloud.githubusercontent.com/assets/3201257/13550093/82c9ae2a-e2e3-11e5-8440-b77cbddf460e.png)

drphilmarshall:
Cooooooool! Nice work @wmwv :-) I'm going to need many of these panels
unpacking for me, I guess others might be in the same boat. Could I please
ask you to find a good place to put them in our Twinkles session slides
here?:

https://docs.google.com/presentation/d/1MdGGDrITW4-n04goJNYBAoVwnQBjRkibxuJY8EJZWXc/edit#slide=id.p

Real quick: how many visits went into these plots? Which dataset are you
working from?

The script that made these, by calling the validate_drp tools - is it
[...]

PR #173, opened 2016-03-11
update bin/plot_point_mags.py to do multibands

jchiang87:
This supersedes `bin/plot_point_mags_gri.py`.  This operates on an output data repository without having to specify the bands or visits as it finds them from the registry.sqlite3 file:
```
$ plot_point_mags.py --help
usage: plot_point_mags.py [-h] [--skip_fit] data_repo outfile

For an output repository of Level 2 data, make a plot of stdev vs median flux
of forced sources.

positional arguments:
  data_repo   Output repository for Level 2 analysis
  outfile     Filename of the output png file

optional arguments:
  -h, --help  show this help message and exit
  --skip_fit  Skip the fitting of the snr curve
```


drphilmarshall:
Sorry for the delay, @jchiang87 - who should review this code? @SimonKrughoff ?

jchiang87:
Sure, Simon would be the right person.  I'm wondering if this sort of script isn't superseded by the `validate_drp` stuff at #167.  Also, since this file gets the data from the forced source FITS catalogs, it doesn't scale very well for large processing runs.   It might make more sense to have it do db queries instead (assuming we get that going as a regular feature of the pipeline output).

drphilmarshall:
Yeah - but we should merge this first and then deal with the db upgrade later. I made a note in #167 for @wmwv to think about.

SimonKrughoff:
All these changes look very good improvements on the original.  Thanks for the excellent improvements @jchiang87 .  I assume I should leave it to you to deal with the merge conflicts merge and close, but I'm open to revising that process.

PR #178, opened 2016-03-11
Student Empowerment

humnaawan:
Check it out, @sethdigel ! We have got the cross validation working for the KNN model, but its still not all that great. So, here's what we're going to try and do:

* [x] Take log of CPU time - this should make KNN work better, Tom predicts :-)
* [x] Try Random Forest - which should Trump(TM) KNN, Phil predicts :-)
* [x] Clean up all notes in notebook
* [x] Remove quasar crap
* [ ] Write a very brief summary of our findings, and recommend how to take what we've done and make it work in production.
* [ ] Figure out binder, so that we can show the notebook live in the wrap-up @drphilmarshall

### DIFF
--- a/examples/notebooks/Run1CPUTimes.ipynb
+++ b/examples/notebooks/Run1CPUTimes.ipynb
@@ -358,11 +358,11 @@
     "\n",
     "# And we'll need a cv iterator:\n",
     "from sklearn.cross_validation import ShuffleSplit\n",
-    "shuffle_split = ShuffleSplit(len(X), 10, test_size=0.4)\n",
+    "shuffle_split = ShuffleSplit(len(X_scaled_knn), 10, test_size=0.4)\n",
     "\n",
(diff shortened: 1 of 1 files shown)


humnaawan:
Alright! Tom and Phil were right - using log(CPUtime) gives a KNN score of 0.81 compared to 0.60, and Random Forest beats KNN by 0.91 to 0.81. Check it out:

rfbeatsknn

humnaawan:
rf predicted vs actual time

humnaawan:
Things that came up:
- Calculate statistics on the accuracy of the Random Forest predictions. 
- Run Grid Random Forest calculation using parameters other than n_estimators.
- Need to account for different number of runs on different hosts. Weight the training sample using the hostname?
- Incorporate Opsim columns as features? Possibly to ensure that the data common to Opsim and PhoSim is consistent.
- Look at finseeing vs. raw seeing.

PR #184, opened 2016-03-23
Issue/58/forced source db ingest

jchiang87:
Hey @jbkalmbach @rbiswas4 !

I was keen to get the Run1 forced source data into MySQL tables so that I could sort out what queries we will be making with qserv and to start playing around with the data.  So over the weekend, I put together the code in this branch, and I ran the ingest on the forced source catalogs of the 985 Run1 visits, using some MySQL tablespace I have on a server in KIPAC.  That ingest took a couple days (we should be able to speed that up though), but once the data were in the tables, querying for the forced source fluxes was very fast, so I wrote a couple classes to serve up light curves based on the Monitor code:

https://github.com/DarkEnergyScienceCollaboration/Twinkles/blob/issue/58/forced_source_db_ingest/python/desc/twinkles/lightCurveFactory.py

and I put it to work in a study on differentiating SN light curves from AGNs:

https://github.com/jchiang87/desc_projects/blob/master/twinkles/ExploringTwinklesRun1.ipynb

That notebook runs in a few minutes.

I'm hoping that the db-based light curve code can eventually go into the Monitor package.   We'll need to have a scalable MySQL db service set up to make use of it though.

Let me know what you guys think.


rbiswas4:
This is great to have this data in a sql database and yes, I agree this is the kind of thing we would like to have in Monitor. I have some more specific comments, which I will put in sometime today. The quick question: I assume you authenticate via your profile, is there a way I could run this code and query the database from here?

jchiang87:
@rbiswas4 I'll need to check on how to get access to those tables for you.  At a minimum, you'll need a SLAC linux account.   Once I have your userid, I'll contact the person administering that server (@slm23) to set up your credentials.

jbkalmbach:
This is really great and the notebook is really interesting. Can you explain more about the changes you made to adding the bandpasses to the registry for sncosmo? I don't quite understand why you need the des and sdss bands as templates?

Also, do you mind if I take the non-database dependent parts of your updated monitor-like code and update the monitor with them now? Finally, do you mind if I point to this when I talk about the monitor during the SSim telecon tomorrow?

jchiang87:
@jbkalmbach Sure, feel free to copy any bits over to Monitor that seem useful and to point to this stuff tomorrow at SSims.

Regarding the bandpassses, the problem I was finding was that the sncosmo code would only fit bands that were covered by the model.  All of the LSST throughput data are defined from 300 nm to 1200 nm, so none of them were covered by the model I was trying to use (salt2).   The relevant lines of code in sncosmo appear to be [here](https://github.com/sncosmo/sncosmo/blob/70a0b356e7c30dfdc2243099ca9516388f7dc4c4/sncosmo/models.py#L1334) where `b.wave[0]` and `b.wave[-1]` would be 300nm and 1200nm, respectively.  So the `overlap` always ends up being `False`.  I tried thresholding the throughput values at 2e-4, but that still caused problems,.  So since I knew the sdss and des bands must already work, I used their ranges to limit the lsst ranges.   In the end, I think we just need to have the lsst throughput files themselves contain the appropriately narrow wavelength ranges.

rbiswas4:
@jchiang87 

Yes, that has been a problem. I think the way we have been doing this is to replace 'salt2' by 'salt2-extended'. This is of course not 'correct' but neither are the simulations, and a more principled approach would emerge from the photometry bake-off. 

Also, do you think we could run your ingest script and create that mysql db locally so that we can all try it out?

jchiang87:
> Also, do you think we could run your ingest script and create that mysql db locally so that we can all try it out?

Definitely.  I'm hoping that `bin/ingest_forced.py` just works out of the box,  but you'll also need to run a modified version of 

https://github.com/jchiang87/desc_projects/blob/master/twinkles/ingest_ref_catalog.py

and if you want the Chisq table, this script:

https://github.com/jchiang87/desc_projects/blob/master/twinkles/chisq_dists.py

I don't think I have a separate script for generating the CcdVisit table, but to do that it would just be

```
from desc.twinkles.db_table_access import CcdVisitTable
ccdvisit = CcdVisitTable(db=<db_name>, read_default_file=<file with credentials>)
ccdvisit.ingestRegistry(<full path to registry.sqlite3 file>)
```

Everything except `ingest_forced.py` should just take a couple minutes.

Let me know how it goes!

PS:  I'm planning to do multiple inserts in the `ForcedSourceTable.ingestSourceCatalog` method, which should speed things up, but I'm not sure by how much.  I haven't tried running multiple ingest jobs in parallel, but that would be another approach.

jchiang87:
BTW, my file with the connection info for the MySQL server I'm using has these entries filled
```
[client]
host=
port=
user=
password=
#database=
```
I have a couple different dbs that I use, so I have the `database` entry commented out.

jchiang87:
@rbiswas4 @jbkalmbach 

I just pushed b3864bd with the multiple insert per query change.  This speeds up the ingest by about a factor of three or so.  On the rather old MySQL server I'm using, that corresponds to an ingest time of ~30 seconds per 47k object forced source catalog.

drphilmarshall:
Or 8-9 hours for Run 1, I guess. Thanks Jim!

drphilmarshall:
Eep, something went wrong [over at travis...](https://travis-ci.org/DarkEnergyScienceCollaboration/Twinkles/builds/118630929). Otherwise, can this be merged?

jchiang87:
I think so.  The failure was on the last push on the branch, which doesn't yet have the .travis.yml config.  The run on the PR, which is what would run after the merge, succeeded.  (Although, the db related tests were skipped...I'll fix that after the merge...)

drphilmarshall:
Alright! Let's do this then.

PR #189, opened 2016-03-24
Setup modifications

rbiswas4:
@danielsf @jchiang87 

Here is a setup that seems to work for me (and I thought could work for everyone). I split the setup into a declare.sh and a setup.sh (talking to Scott), where the declare is meant to be run only once by the user. The setup is meant to be run everytime a new shell is opened, *after* setting up lsst-sims in whatever way the user usually does. I did this to avoid the problem of having to specify what tags the user would have on lsst_sims.

### DIFF
--- a/setup/declare.sh
+++ b/setup/declare.sh
@@ -1,0 +1,2 @@
+# Run this the first time only
+eups declare -r . Twinkles $USER  --tag=$USER
--- a/setup/setup.sh
+++ b/setup/setup.sh
@@ -1,9 +1,2 @@
-inst_dir=$( cd $(dirname $BASH_SOURCE)/../..; pwd -P )
-if [ ! -d "$inst_dir/ups_db" ]; then
-   mkdir $inst_dir/ups_db
-fi
-export EUPS_PATH=$inst_dir:${LSSTSW}/stack
-. ${LSSTSW}/bin/setup.sh
-setup lsst_sims
-eups declare Twinkles -r . -c
-setup Twinkles
+# Setup lsst_sims in the way you usually do
+setup Twinkles -t $USER
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -19,7 +19,7 @@ this is becomes your default version of python.
 
 First you need to add the LSST "channel:"
 ```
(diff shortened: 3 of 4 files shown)


setup/setup.sh
```diff
@@ -1,9 +1,2 @@
...
-if [ ! -d "$inst_dir/ups_db" ]; then
-   mkdir $inst_dir/ups_db
-fi
-export EUPS_PATH=$inst_dir:${LSSTSW}/stack
-. ${LSSTSW}/bin/setup.sh
-setup lsst_sims
-eups declare Twinkles -r . -c
-setup Twinkles
+# Setup lsst_sims in the way you usually do
+setup Twinkles -t $USER
```
danielsf: This does not work if the user does not have a version of lsst_sims tagged $USER.

I tried running this with a version of eups that only knew about the -t sims version of lsst_sims, it came back with
```
setup: in file /Users/danielsf/physics/lsst_160212/Development/garage/Twinkles/ups/Twinkles.table: Product lsst_sims not found
```
this command
```
setup Twinkles -t $USER -t sims
```
worked.

I think we are going to have to settle for writing a very good README that explains to users how to navigate EUPS, rather than scripting everything for them.  There are too many possible permutations (which is, I guess, why EUPS exists).

setup/declare.sh
```diff
@@ -0,0 +1,2 @@
+# Run this the first time only
+eups declare -r . Twinkles $USER  --tag=$USER
```
danielsf: This is going to fall prey to the bug outlined here

https://community.lsst.org/t/change-of-behavior-in-eups/523/7

safer to do
```
eups declare -r . Twinkles $USER
eups declare -r . Twinkles $USER --tag=$USER
````
as Russel explains in that community post (otherwise, on EUPS 2.0.1, the tag will get blown away every time EUPS is re-setup; this is a bug that will be fixed in the next EUPS release).

drphilmarshall:
I'm going to discreetly close this PR without merging, as things have moved on a lot since this attempt to streamline the stack setup got stuck in the mud, like so many others before it and since.

PR #196, opened 2016-03-29
Added in new lens catalog. Updated sprinkler with new lens catalog filename.

### LINKED ISSUE #192: Add in remaining TDC1 rung 4 lenses.
In #117 we decided to sprinkle in a set of lenses defined by the TDC1 rung 4, but for run 1 we settled on only sprinkling in the first 200 from that rung. We need to update our sprinkling catalog with the rest of rung 4 since only taking the first 200 left out all quads from the lens catalog.

jbkalmbach:
Now we have quads. Fixes #192


drphilmarshall:
W00t!

PR #212, opened 2016-04-07
Adding module that produces a subset of the data repository containin…

rbiswas4:
script to create subsets of the data repository for smaller size.
	new file:   copyData.py

### DIFF
--- a/bin/copyData.py
+++ b/bin/copyData.py
@@ -1,0 +1,82 @@
+"""
+Copies parts of a data repository corresponding to certain visits into a new
+repository to make a small test repository
+
+To Use change the strings in script part:
+    - change the srcdir to point to the repo that is to be copied
+    - obtain the list visits to a list of obshistIDs of visits that are desired in the subset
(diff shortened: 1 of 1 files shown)


rbiswas4:
See the issue described in #213

drphilmarshall:
Thanks @rbiswas4 ! This looks reasonable, but there's no demo notebook or markdown doc describing its use. Could be a good idea, if the script is to be run at SLAC or NERSC by someone else?

drphilmarshall:
@rbiswas4 What do you need here? Someone else to try and use the code, and document it in the process? Is this something that could be done as part of one of the LSST DESC Note write-ups?

rbiswas4:
@drphilmarshall Sorry, noticing this after a while. There is an example script using this at the bottom of the module (after the `if __name__=='__main__'` part). Should we make it a notebook?

drphilmarshall:
That's a good idea! Should be quick to make and test, no? Thanks!

PR #227, opened 2016-04-21
Add --output argument to commands in Cookbook

wmwv:
An --output repo is now required since DM-4692


wmwv:
@SimonKrughoff   I added --output specifications to the examples in the Cookbook.md.

Could you double-check that you agree with these changes?

PR #237, opened 2016-05-06
Added new fits table. Adjust change in galtile id so now we can uniqu…

jbkalmbach:
…ely id up to 10000 lens images.


jbkalmbach:
If nobody has any comments, I'll go ahead and merge this in a couple hours.

PR #240, opened 2016-05-12
Issue/186/cpu time predictor

sethdigel:
This adds the Run 1 metadata, a script to generate a random forest predictor of the CPU time, and a class to use that predictor to estimate the CPU time for a given (filter, moonalt, moonphase), or ObsHistID from the kraken_1042 Opsim run.


python/desc/twinkles/phosim_cpu_pred.py
```diff
@@ -0,0 +1,91 @@
...
+            moonphase_arg]]))
+    def opsimdb(self, obsid):
+        rec = self.factory.create('obsHistID filter moonAlt moonPhase'.split(), 
+            'Summary', 
+            condition="where obsHistID='%i' order by obsHistID asc"%obsid)
+
+        if rec.size <> 0:
+            # Translate the filter string into an index 0-5
+            filtmap = 'u g r i z y'.split()
+            filt = filtmap.index(rec["filter"].values[0])
```
jchiang87: The following might be more idomatic python and easier to understand:
```
filt = 'ugrizy'.find(rec['filter'].values[0])
```

python/desc/twinkles/phosim_cpu_pred.py
```diff
@@ -0,0 +1,91 @@
...
+    RF_pickle.p is written by run1_cpu_generate_rf.py
+    """
+    def __init__(self, rf_pickle_file='RF_pickle.p', opsim_db_file='/nfs/farm/g/lsst/u1/DESC/Twinkles/kraken_1042_sqlite.db'):
+        self.RFbest = pickle.load(open(rf_pickle_file, 'rb'))
+        self.factory = SqliteDataFrameFactory(opsim_db_file)
+    def __call__(self, filter_arg, moonalt_arg, moonphase_arg):
+        return 10.**self.RFbest.predict(np.array([[filter_arg, moonalt_arg, 
+            moonphase_arg]]))
+    def opsimdb(self, obsid):
+        rec = self.factory.create('obsHistID filter moonAlt moonPhase'.split(),
```
jchiang87: It will probably be more efficient to create the data frame for all obsHistIDs in the constructor, e.g., with
```
factory = SqliteDataFrameFactory(opsim_db_file)
self.obs_conditions = factory.create('obsHistID filter moonAlt moonPhase'.split(), 'Summary')
```
then in `.opsimdb(obsid)`, do
```
rec = self.obs_conditions[self.obs_conditions['obsHistID'] == obsid]
```
since sqlite is generally slow whereas pandas is supposed to be fast for these sorts of selections.

python/desc/twinkles/phosim_cpu_pred.py
wmwv: I would say just go ahead and be explicit.  Use variable names to help explain to the reader what is a filter index and what is a filter name.

```
filter_name_to_index = {'u': 0, 'g': 1, 'r': 2, 'i': 3, 'z': 4, 'y': 5}
filter_name = rec["filter"].values[0]]
filter_index = filter_name_to_index[filter_name]
[...]
return filter_index, moonalt, moonphase
```

drphilmarshall: I agree with @wmwv. Jim hates us right now

jchiang87: It's fine Phil.  pylint will sometimes complain about too many local variables (which is itself annoying, but the acceptable number is configurable), so I've tried to be more concise recently, relying on common idioms (common in my estimation, at least) to provide clarity.  Being very explicit in variable usage is fine.

drphilmarshall: Ah, good :-)
I have not yet noticed what pylint makes of my "literate python" style,
which aims to construct python statements that read as grammatically
correct English.

python/desc/twinkles/phosim_cpu_pred.py
sethdigel: Thanks, Jim.  I've implemented this in the next commit, also at Jim's suggestion adding a 'condition' argument to factory.create to select only the visits within the Twinkles Run 1 region.

python/desc/twinkles/phosim_cpu_pred.py
sethdigel: Thanks.  In the next commit I went with something like Jim's recommendation - aiming for concise clarity - so maybe more lyrical than literate.

bin/run1_cpu_generate_rf.py
```diff
@@ -0,0 +1,90 @@
+"""
```
jchiang87: In order for this to be a standalone executable, the first line of this script should be, e.g., 
```
#!/usr/bin/env python
```

sethdigel: Ah, yes.  Thanks.  Done.

bin/run1_cpu_generate_rf.py
```diff
@@ -0,0 +1,91 @@
...
+Generates Random Forest predictor for Run 1 CPU times, based on the iPython
+ML notebook by Phil Marshall
+https://github.com/drphilmarshall/StatisticalMethods/blob/master/examples/SDSScatalog/Quasars.ipynb
+and updates by Humna Awan and Tom McClintock
+https://github.com/DarkEnergyScienceCollaboration/Twinkles/blob/master/examples/notebooks/Run1CPUTimes.ipynb
+
+@author: sethdigel
+"""
+from __future__ import print_function
```
jchiang87: The `pylint --py3k` checks want this line to include `absolute_import`, i.e.,
```
from __future__ import print_function, absolute_import
```

python/desc/twinkles/phosim_cpu_pred.py
```diff
@@ -0,0 +1,103 @@
+"""
+Estimates the CPU time required for a phosim simulation of a 30-s visit.  The
+inputs are filter, moonalt, and moonphase, or obsHistID (an Opsim ID from
+a specified (hard coded) Opsim sqlite database.
+
+The random forest is generated (and saved as a pickle file) by 
+run1_cpu_generate_rf.py, using only the filter, moonalt, and moonphase
+features.  This script needs to be run only once.
+"""
+
```
jchiang87: For python 3 compatibility, please add
```
from __future__ import print_function, absolute_import
```

python/desc/twinkles/phosim_cpu_pred.py
```diff
@@ -0,0 +1,103 @@
...
+        return self.cputime(filter_index, moonalt, moonphase)
+
+    def conditions(self, obsid):
+        """
+        Return the relevant observing conditions (i.e., those that are
+        inputs to the Random Forest predictor) for the given obsHistID
+        """
+        rec = self.obs_conditions[self.obs_conditions['obsHistID'] == obsid]
+
+        if rec.size <> 0:
```
jchiang87: `<>` is apparently deprecated in python 3 in favor of `!=`

bin/run1_cpu_generate_rf.py
sethdigel: Belated thanks, Jim.  I somehow lost track of this.  I've made the changes now that you recommended.

PR #245, opened 2016-05-20
Sims Recipe

drphilmarshall:
Reorganized doc/Cookbook to include Sims Recipe, fixed links etc. I think this is ready for @danielsf and @rbiswas4 to take over.


drphilmarshall:
Note that I moved the phosim data generation part of the original cookbook into a new `Sims_Recipe.md` file, but I left the reference catalogs for calibration part in the DM Level 2 recipe (now with a new filename!).

PR #259, opened 2016-06-10
add function to create in-memory image hdus from Exposures

jchiang87:


tests/test_PostageStampMaker.py
```diff
@@ -90,5 +95,27 @@ def test_create_sequence_function(self):
...
+        my_hdu = convert_image_to_hdu(my_stamp)
+
+        # Write the Exposure to a FITS file to test the image data and
+        # WCS keywords in the in-memory HDU against the versions
+        # written by the Stack.
+        stamp_file = 'temp_stamp.fits'
+        my_stamp.writeFits(stamp_file)
+        stack_hdu = fits.open(stamp_file)[1]
+
+        for value1, value2 in zip(my_hdu.data.flatten(),
```
SimonKrughoff: Note that numpy has an `array_equal` function, so you could say: `self.assertTrue(numpy.array_equal(my_hdu.data, stack_hdu.data)`.

Another option is to say `self.assertTrue((my_hdu.data==stack_hdu.data).all())`

jbkalmbach:
@jchiang87. What do you think about adding moving your postage stamp code to the monitor? Otherwise, the Monitor will depend on Twinkles for image visualization. Since I think we want the collaboration to use tools like the Monitor to analyze data and light curves should we collect such tools there instead of in the Twinkles repo which is more focused?

jchiang87:
@jbkalmbach Sure, we could do that, although it might make more sense just to have a plotting/display package that has all of the general plotting code in it.   I'll look into moving the postage stamp-related code to Monitor for now.

jchiang87:
@jbkalmbach I just added this code to Monitor.  I'll delete it from this Twinkles branch.

jchiang87:
I've moved this code to the Monitor package

PR #272, opened 2016-07-01
First stab at Weekly Summary

heather999:
First attempt - purely to start get something out, with much room for improvement :)
Modified the index.html (forked version here: https://heather999.github.io/Twinkles/) to include a link to a new weekly/index.html (you can see what it looks like here: https://heather999.github.io/Twinkles/weekly/index.html) which just points to our first weekly summary: https://heather999.github.io/Twinkles/weekly/2016/06/weekly_20160630.html

For now the weeklies would be organized under: weekly/2016/Month

The HTML that Google Docs puts out is rather ugly, I'll likely use pandocs in the future to format it better... and the weekly/index.html needs to be prettied up.


drphilmarshall:
@heather999 Nice, the index looks much better in style - thank you! Looking at your headline, I realized that as part of the notes clean up I could insert a headline myself, below the meeting details (which don't need to be redistributed). I made it in paragraph style, not heading, so that it doesn't screw up the table of contents, but hopefully it'll pandoc well into HTML! Check it out:

image

heather999:
Ah yes that does look good @drphilmarshall .  pandocing now - and I need to do something about the image files.  Pandoc extracts them and references them in the HTML - so I'll just have to store them separately. 
I'll apply the same theme to the weeklies themselves so the look will be consistent.
I'll also write a recipe for what I did so I can do this again next time :)

drphilmarshall:
Great, thanks Heather! I wondered what would happen to the images... I
think we do want to include them, though, as they'll make for a much nicer
weekly summary, don't you think?

heather999:
July 7th weekly had been added and can currently be viewed here:
https://heather999.github.io/Twinkles/weekly/2016/07/weekly_20160707.html

heather999:
added a doc/howto.rst but in the gh-pages branch GitHub doesn't seem to render it as it does on the main branch.  Perhaps I should move the doc up to the main branch.

drphilmarshall:
Great stuff, @heather999 ! I don't see the `howto.rst` yet - maybe we need a `Website` subfolder in docs to keep website notes in, what do you think? Then your weeklies howto might be better named weeklies_howto.rst. 

Other than this, I think this is looking very good. Is it hard to get the weekly posts into the standard site style? That'd finish things off nicely I think. Great to see the images in there! :-)

heather999:
and now getting back to documenting this. I'll have to make a separate pull request to handle updating/adding a Doc/Website directory as that will ultimately go onto master and all of the above belongs on the gh-pages branch.  I'll take care of that shortly.

drphilmarshall:
I see the howto! Very nice. Thanks @heather999 ! :-)

PR #275, opened 2016-07-09
Issue/160/installation - Heather instructing, Phil testing

drphilmarshall:
I am following the [instructions from Heather](https://github.com/DarkEnergyScienceCollaboration/Twinkles/blob/issue/160/installation/doc/Setup/Getting_started_on_your_laptop.rst), and editing them as I go. Here's the growing list of feedback items I am generating, to be addressed:

* [x] Instructions are mac-specific, I point this out

* [x] `conda create  --name lsst python=2` generates two (apparently non-destructive) errors: 
`SSL verification error: [SSL: UNKNOWN_PROTOCOL] unknown protocol (_ssl.c:590)`

* [x] `conda create` does not complete fully, for me: 
`No psutil available.
To proceed, please conda install psutil#` I did this before carrying on, with apparently no ill effects.

* [x] `source activate lsst` only works if you are in `${HOME}/miniconda/bin` - I changed the instructions to use the full path

* [x] The `activate` command fails, with error
`BASH_VERSION: Undefined variable.`
I am a C-shell user. I had to switch to `bash` to complete the installation - I added notes on this.

* [x] Maybe we should recommend `bash` further up? I need to check that things work when I switch back to `tcsh` though... and they don't. Instructions are now correct regarding use of c-shell.

* [x] The PhoSim test fails if you don't have access to the CatSim database at UW. This is a pretty stringent test, but hey - that's what it takes to run Twinkles on your laptop. I added notes on this.

### DIFF
--- a/doc/Setup/Getting_started_on_your_laptop.rst
+++ b/doc/Setup/Getting_started_on_your_laptop.rst
@@ -38,18 +38,23 @@ Pipelines <https://pipelines.lsst.io/install/conda.html>`_. The first thing that
    conda update conda
 
 The following commands will download and activate the current release versions of the LSST Science Pipelines in a 
-new Conda environment named "lsst":
+new Conda environment named "lsst". At various times they need to to type `y`, so unfortunately you cannot leave them too long. They 
+each take a few minutes, except for the `conda install`, which takes a bit longer.  
 
 .. code-block:: bash
 
    conda config --add channels http://conda.lsst.codes/stack  
    conda create --name lsst python=2
+   
+You are now ready to install the `lsst` software. If you are a C-shell user, the `source activate` command below
+will not work, as that script only supports `bash` and `zsh`. A workaround is to switch to a `bash` shell at this point, 
+and then stay in that bash shell to do the other two commands below.
+
    source ${HOME}/miniconda2/bin/activate lsst
    conda install lsst-distrib lsst-sims
    source eups-setups.sh
 
-
-You are now set up to use the DM Stack in the current shell.
+You should now be set up to use the DM Stack in the current `bash` shell.
 
 Set Up Environment to Use LSST Code
 -----------------------------------
--- a/doc/Setup/Getting_started_on_your_laptop.rst
+++ b/doc/Setup/Getting_started_on_your_laptop.rst
@@ -85,13 +85,21 @@ Install Optional Python Modules Not Included with DMStack
     
 Install PhoSim
 -----------------------
-The PhoSim Confluence page is available `here <https://confluence.lsstcorp.org/display/PHOSIM>`_.
+The PhoSim Confluence page is available `here <https://confluence.lsstcorp.org/display/PHOSIM>`_. The code is available 
+from LSST via `git`, and needs the `cfitsio` and `fftw3` libraries: you'll be asked to point to their locations by the `PhoSim` 
+configure script, or if you can't, it will offer to install them for you from source.
      
 .. code-block:: bash
 
     mkdir ~/repos
     cd ~/repos
     git clone https://stash.lsstcorp.org/scm/sim/sims_phosim.git
+
+This takes a few minutes, as the `sims_phosim` repo is large. Once it has been downloaded,
+ 
+.. code-block:: bash
+
+    cd sims_phosim
     setup cfitsio
     setup fftw
     ./configure
--- a/doc/Setup/Getting_started_on_your_laptop.rst
+++ b/doc/Setup/Getting_started_on_your_laptop.rst
@@ -39,7 +39,7 @@ Pipelines <https://pipelines.lsst.io/install/conda.html>`_. The first thing that
 
 The following commands will download and activate the current release versions of the LSST Science Pipelines in a 
 new Conda environment named "lsst". At various times they need to to type "`y`", so unfortunately you cannot leave them too long. They 
-each take a few minutes, except for the `conda install`, which takes a bit longer.  
+each take a few minutes, except for the `conda install`, which takes about an hour.  
 
 .. code-block:: bash
 
@@ -80,7 +80,7 @@ Install Optional Python Modules Not Included with DMStack
 .. code-block:: bash
 
     conda install nose
-    conda install coverge
+    conda install coverage
     conda install iminuit
     
 Install PhoSim
--- a/doc/Setup/Getting_started_on_your_laptop.rst
+++ b/doc/Setup/Getting_started_on_your_laptop.rst
@@ -51,8 +51,11 @@ new Conda environment named "lsst":
 
 You are now set up to use the DM Stack in the current shell.
 
-Environment Set up to Use LSST Code
--------------------------------------
+Set Up Environment to Use LSST Code
+-----------------------------------
+Every time you start a new shell, you need to set up its environment to enable use of the LSST code. The following lines could be 
+worth adding to your `.bashrc` file or equivalent.
+C-shell users would use `eups_setup.csh` and `setenv` in their `.login` file.
 
 .. code-block:: bash
 
@@ -63,10 +66,6 @@ Environment Set up to Use LSST Code
    setup obs_lsstSim
    setup lsst_sims
 
-This needs to be done every time you start a new shell - so these lines 
-could be worth adding to your `.bashrc` file or equivalent. Note that 
-c-shell users can use `eups_setup.csh` and use `setenv` in their `.login` 
-file.
 
 Install Optional Python Modules Not Included with DMStack
 ----------------------------
@@ -129,6 +128,7 @@ This is where the objects that will populate the catalog are stored.
 
 Supernova Sprinkling Setup
 ---------------------------
+Coming soon!
 
 
 Important Installation Notes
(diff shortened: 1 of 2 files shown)


heather999:
As I'm more of a Linux user - I'm happy to expand the directions to include the linux-specifc bits.  I'm guessing it will be clearer if I just duplicate the "Apple laptop" instructions as a 2nd section in to doc.  

I'm a little surprised that the having your miniconda installation in your path isn't enough when calling "source activate lsst" on the Mac, rather than providing the full path.  I'm upgrading my old Mac to El Capitan right now - so I can give it a fresh trial, going from what you have there now.

The psutil error surprises me too..  I haven't seen that at

drphilmarshall:
It seems as though the source command works differently in `bash` and `tcsh`. In `tcsh` you have to provide the correct full path to the file you want to source; in `bash` I think it searches your path, or perhaps the tree beneath you? Not sure. This is not the first time I have come across `bash` being assumed - perhaps we shoudl recommend that all DESC coders switch to the `bash` shell?

The `psutil` error does not seem too serious, I'll check it off. Now working on `PhoSim` (sp?)

heather999:
I think it would be best to pick a shell and stick with it for the examples and perhaps offer suggestions for other shells, particularly C-shell.  In the past, bash worked best when playing with DMstack and I've tended to just stick with that.  I don't know what Mac users tend to do - though perhaps C-shell is more commonly used there?  If we're going to have instructions per OS, perhaps we can tailor the shell accordingly if that is more appropriate for that user community.
I should note - I found a blurb in the NERSC doc recommending c-shell, but I always default to bash in their environment and haven't run into any problems there.  Note to self: find that nersc web page again..

rbiswas4:
@heather999 I don't think the shell is a linux-vs-mac users question. The default shell in OSX is also bash. So, i am guessing you will have different shell users in different OS.

drphilmarshall:
`PhoSim` test failure! Looks like we have a script that is expecting LSST software in `~/.lsst`, whereas actually it is in `~/miniconda2/envs/lsst` (or something). Here's the command and error message:

```bash
(lsst) bash-3.2$ python $SIMS_CATUTILS_DIR/examples/generatePhosimInput.py
sh: sysctl: command not found
sh: sysctl: command not found
sh: sysctl: command not found
/Users/pjm/miniconda2/envs/lsst/opt/lsst/sims_catalogs_generation/python/lsst/sims/catalogs/generation/db/dbConnection.py:438: UserWarning: Duplicate object type id 25 specified: 
Output object ids may not be unique.
This may not be a problem if you do not want globally unique id values
  'want globally unique id values')
/Users/pjm/miniconda2/envs/lsst/opt/lsst/sims_catalogs_generation/python/lsst/sims/catalogs/generation/db/dbConnection.py:438: UserWarning: Duplicate object type id 40 specified: 
Output object ids may not be unique.
This may not be a problem if you do not want globally unique id values
  'want globally unique id values')
sh: sysctl: command not found
sh: sysctl: command not found

For more information on database authentication using the db-auth.paf policy file see: https://confluence.lsstcorp.org/display/SIM/Accessing+the+UW+CATSIM+Database

Traceback (most recent call last):
  File "/Users/pjm/miniconda2/envs/lsst/opt/lsst/sims_catUtils/examples/generatePhosimInput.py", line 18, in <module>
    obsMD = OpSim3_61DBObject()
  File "/Users/pjm/miniconda2/envs/lsst/opt/lsst/sims_catUtils/python/lsst/sims/catUtils/baseCatalogModels/OpSim3_61DBObject.py", line 66, in __init__
    super(OpSim3_61DBObject, self).__init__(driver=driver, host=host, port=port, database=database)
  File "/Users/pjm/miniconda2/envs/lsst/opt/lsst/sims_catalogs_generation/python/lsst/sims/catalogs/generation/db/dbConnection.py", line 274, in __init__
    port=self.port, verbose=self.verbose)
  File "/Users/pjm/miniconda2/envs/lsst/opt/lsst/sims_catalogs_generation/python/lsst/sims/catalogs/generation/db/dbConnection.py", line 113, in __init__
    self._connect_to_engine()
  File "/Users/pjm/miniconda2/envs/lsst/opt/lsst/sims_catalogs_generation/python/lsst/sims/catalogs/generation/db/dbConnection.py", line 121, in _connect_to_engine
    authDict = {'username': DbAuth.username(self._host, str(self._port)),
  File "/Users/pjm/miniconda2/envs/lsst/opt/lsst/daf_persistence/python/lsst/daf/persistence/persistenceLib.py", line 986, in username
    return _persistenceLib.DbAuth_username(*args)
lsst.pex.exceptions.wrappers.RuntimeError: 
  File "src/DbAuth.cc", line 83, in std::pair<std::string, std::string> search(const std::string &, const std::string &)
    /Users/pjm/.lsst directory is missing or accessible by others {0}
lsst::pex::exceptions::RuntimeError: '/Users/pjm/.lsst directory is missing or accessible by others'
```

rbiswas4:
@drphilmarshall 

 I think the instructions should include those for creating that .lsst directory. This is a directory that holds credentials for remote authentication to UW databases:  The instructions for doing so are at:
https://confluence.lsstcorp.org/display/SIM/Accessing+the+UW+CATSIM+Database, but need some communication with people (so one should expect some time lag, and that should perhaps be in the beginning of the instructions as a warning).  Some parts of that UW+CATSIM page are password protected too, so that information should not be put elsewhere.

drphilmarshall:
Yes, that sounds important, @rbiswas4 ! Interesting that our basic PhoSim test requires having this access - it seems like a high bar, but I guess if the goal is to be able to "run Twinkles on your laptop" then the user needs those credentials...

rbiswas4:
>  Interesting that our basic PhoSim test requires having this access - it seems like a high bar
@drphilmarshall Obviously, if the intent is to test that the user's install of phosim is correct, this is not necessary. I believe phosim itself comes with a set of tests which should be sufficient for this purpose. Moreover, if you look at those lines of code:
The first step accessed the database and produced a input file, which was used by phosim in the second step. One could easily check the input file into the data directory and have phosim use that as an example, so that users can ensure that their phosim is up and running. Maybe we can split this in the instructions?

Of course, if you did really want to generate images, you really do want access to the database, and you do have to get past the 'high bar'.

rbiswas4:
@drphilmarshall Also, are you noticing that travis is often failing with your pushes? Since you are actually adding documentation, does this mean that the wait time has not solved the travis install issue?

doc/Setup/Getting_started_on_your_laptop.rst
```diff
@@ -0,0 +1,179 @@
...
+   source activate lsst
+   conda install lsst-distrib lsst-sims
+   source eups-setups.sh
+
+You should now be set up to use the DM Stack in the current `bash` shell.
+
+Set Up Environment to Use LSST Code
+-----------------------------------
+Every time you start a new shell, you need to set up its environment to enable use of the LSST code. The following lines could be 
+worth adding to your `~/.bash_profile` file or equivalent.
```
danielsf: Historically, adding LSST setup commands to ~/.bash_profile has caused problems for people.  Setting up the LSST stack usually involves using a specific installation of python, which is not always the user's default preferred python installation.  I would instead recommend that they copy those commands to an ```setUpLSST.sh``` script, and just ```source``` that script whenever they want to run the LSST stack.

rbiswas4:
@heather999, @drphilmarshall, @danielsf 

This is about the installation instructions here for the conda build and the issue is about having
 
1. both DM and Sims stacks work on the same installation 
2. using updates.
  
I had a short discussion about the installation instructions with @danielsf as I was not entirely sure about this. The current installation uses the LSST releases. So 1. should always work! However, these releases have a cadence of either 1 or 2 a year. This means that in the intermediate time, there will be updates to the DM and definitely the SIMs stack that will be essential for Twinkles users. This was the rationale between Scott creating tagged releases of both DM and sims at a much higher cadence. The instructions for installing these release are at https://confluence.lsstcorp.org/display/SIM/Catalogs+and+MAF. The main change for these is the conda channel. @danielsf is the expert to talk to and can tell you more about the cadence of this set.

heather999:
@rbiswas4 Thank you - this bit has been on my mind actually.  The current instructions are a bit of a simplification but a step forward.  I'd like to take some time to make sure we document this well.  Are you or @danielsf at the DESC meeting at Oxford this week?  I'm just thinking it wouldn't hurt to chat for 15 minutes - but I can also just start from the Confluence page you referenced above and go from there!

drphilmarshall:
I'm thinking we should issue Rahul and Scott's suggestion separately, and
merge this for now. Talking this week is a very good idea - but I'm not
sure either are here in Oxford?

rbiswas4:
@heather999 
Unfortunately neither @danielsf or I are at Oxford this week. The current structure should be fine at least for Oxford since I don't think the conda sims channel would give you anything new at the moment.

heather999:
@drphilmarshall any objection to my deleting this branch now that it's been merged into master?  I figure it's a good habit to get into, especially for doc updates.

drphilmarshall:
Nope, please do go ahead and delete it - it's served its purpose. Thanks!

PR #303, opened 2016-08-29
Adding some initial thoughts on goals for Twinkles that can be

rbiswas4:
Adding initial thoughts for SN goals that we should try with Twinkles. Not quite ready for merge, but feedback would be good.
new file:   planForTwinkles3.md


drphilmarshall:
Huh - I got confused by the new review UI, sorry! I wanted to *view* my report, and didn't realize that it was already visible as a comment in this thread... I'll see if I can change the status to "changes requested" again...

Issue #316, opened 2016-09-15
Generate reference catalogs with data for all six bands


jchiang87:
This involves modifying [bin/make_reference_catalog.py](https://github.com/DarkEnergyScienceCollaboration/Twinkles/blob/master/bin/make_reference_catalog.py) to include u, z, and y bands, and then using the resulting reference catalog to generate the astrometry.net index files (as [documented in the Level 2 cookbook](https://github.com/DarkEnergyScienceCollaboration/Twinkles/blob/master/doc/Cookbook/DM_Level2_Recipe.md)) with band-specific fluxes for each band.  This will help ensure the zero points found in the Level 2 processing are correct for each band.  See #267.

rbiswas4:
👍 

Incidentally, does this change the few percent change in the other bands that we had been looking at? The change in catsim might also help in fixing that (it was galactic extinction which was being applied wrongly).

jchiang87:
Comparing the magnitudes I've been seeing from CatSim to the ones in the cookbook, there will be differences in the zero points in all bands (from this morning's run, r, g, and i magnitudes also differ by 0.2 mag).   I plan to re-run `processEimage.py` on all of the Run1.1 visits (also turning on the cosmic ray repair) to get new zero points.   I can reload my local pserv db instance with the modified data to show the outcome of these corrections.  If we decide it's desirable, we can also update the Twinkles Run1.1 tables at NERSC (or add a Run1.1_new_zp project to preserve the current data).   I'm not sure it's worth re-running all of Level 2, but it may be something we want to do, depending on the Run3 status.

jchiang87:
The index files are here `/nfs/farm/g/desc/u1/data/PhoSim-Deep-Precursor/and_files/`

PR #317, opened 2016-09-15
update to produce entries for all six LSST bands

jchiang87:


jchiang87:
This pertains to #316

jchiang87:
@SimonKrughoff Could you have a quick look at this please and sign off it seems ok?  btw, the `starnotgal` column seems not to be available from the catsim tables anymore.  I added it by hand before generating the astrometry.net index files.

bin/make_reference_catalog.py
```diff
@@ -21,20 +26,28 @@ class TwinklesReference(InstanceCatalog, AstrometryStars, PhotometryStars):
...
+    stars = CatalogDBObject.from_objid('allstars', host=args.host,
+                                       port=args.port)
 
-    # First get the reference catalog
-    stars = CatalogDBObject.from_objid('allstars')
     while True:
         try:
-            ref_stars = TwinklesReference(stars, obs_metadata=obsMetaDataResults[0])
+            ref_stars = TwinklesReference(stars,
+                                          obs_metadata=obsMetaDataResults[0])
```
SimonKrughoff: This line break doesn't really seem necessary.  You get 110 characters (or is the DESC standard different from LSST?).

bin/make_reference_catalog.py
```diff
@@ -8,10 +13,10 @@
...
 
 class TwinklesReference(InstanceCatalog, AstrometryStars, PhotometryStars):
+    "Reference stars for Twinkles astrometry."
     catalog_type = 'twinkles_ref_star'
-    column_outputs = ['uniqueId', 'raJ2000', 'decJ2000', 'lsst_g',
-                      'lsst_r', 'lsst_i', 'starnotgal',
-                      'isvariable']
+    column_outputs = ['uniqueId', 'raJ2000', 'decJ2000', 'lsst_u', 'lsst_g',
+                      'lsst_r', 'lsst_i', 'lsst_z', 'lsst_y', 'isvariable']
     default_columns = [('isresolved', 0, int), ('isvariable', 0, int)]
```
SimonKrughoff: For the new system we want isresolved, so I think it should be that starnotgal should be changed to isresolved in line 13 rather than removing that column completely.

bin/make_reference_catalog.py
```diff
@@ -21,20 +26,28 @@ class TwinklesReference(InstanceCatalog, AstrometryStars, PhotometryStars):
...
+    obsMetaDataResults = generator.getObservationMetaData(fieldRA=(53, 54),
+                                                          fieldDec=(-29, -27),
+                                                          boundLength=0.3)
+
+    stars = CatalogDBObject.from_objid('allstars', host=args.host,
+                                       port=args.port)
 
-    # First get the reference catalog
-    stars = CatalogDBObject.from_objid('allstars')
     while True:
```
SimonKrughoff: It would be good to document that this is a hack to retry in the case of db connection failure.  I hope there is a better way to deal with this.

jchiang87:
It's fine to remove the `instcat_to_astrometry_net_input` function, but the `build_index_files` function itself is still useful, I think, if only because I can never remember the 10 or so command line options for `build-astrometry-index` command.

bin/make_reference_catalog.py
SimonKrughoff: O.K. no problem.

Issue #297, opened 2016-08-25
Understand the flux dropouts in some of the Run 1.1 light curves


jchiang87:
For a number of objects in our Run 1.1 Level 2 results, we are seeing light curves that look like, e.g., objectId=40514 in the u band:
![objectid_40514_cropped](https://cloud.githubusercontent.com/assets/1994473/17986896/e08a25f0-6ad1-11e6-8233-188babe627b4.png)
The PhoSim centroid files don't show this type of variability for these sources on the input side, so these are artifacts coming out of the Level 2 pipeline.  In order to understand this, I extracted some fluxes for the visit indicated by the red dot and for the visit immediately preceding it in our data.  Here are some results:
```
  visit   PsfFlux (DN, calib)    Ap_3_0 (DN, calib)   Ap_4_5  (DN, calib)    Ap_9_0 (DN, calib)
 703312  3651.2683   1.30e-05  3048.4722   1.09e-05  3392.5642   1.21e-05  3315.2795   1.18e-05  
 704236   227.4048   7.63e-07  2629.8508   8.83e-06  3296.7612   1.11e-05  3802.4734   1.28e-05  
```
The pairs of columns after the visit column are the DN and calibrated flux values for the `base_PsfFlux_flux`, which we are using in the Level 2 tables, and the fluxes for circular apertures of size 3.0, 4.5, and 9.0 pixels (`base_CircularAperture_n_m_flux` in the `forced_src` schema).  The cut-out images for these two visits don't look significantly different, though the brightest pixel is different for each, perhaps indicating a displacement of the image on the focal plane.

At our [Aug 25 weekly meeting](https://docs.google.com/document/d/1NfIO08uqdncbApLkJyAH-UDEpA25lWCU9s26jjdd5jo/edit?pref=2&pli=1#heading=h.p3giulgark0l), I showed these results.   @drphilmarshall and @rbiswas4 suggested looking at the model PSF size, shape, and centroid position, as well as filter and airmass, suspecting that this could be from differential chromatic refraction (DCR) since the effect is most prominent in the u band.

Some questions:
- Should we be using a different flux measurement in the Level 2 tables?
- Do these dropouts indicate that something in the Stack code needs to be fixed?
- What else should be looked at in order to understand these features?

sethdigel:
In case it matters, here are some OpSim parameters for the two visits.  vskybright was essentially the same for both (~21.4 magnitudes/sq arcsec)

[...]

PR #324, opened 2016-09-22
turn on cosmic ray repair

jchiang87:
This addresses #322 

@SimonKrughoff : Is this all we need to turn on to handle cosmic rays?  What about `charImage.repair.doInterpolate`?  Is that relevant for this?

@tony-johnson

### DIFF
--- a/workflows/DM/processEimage
+++ b/workflows/DM/processEimage
@@ -1,3 +1,3 @@
 mkdir -p ${OUT_DIR}
 setup -m none -r ${AND_DIR} astrometry_net_data
-processEimage.py ${IN_DIR} --id visit=${VISIT} --output ${OUT_DIR} --doraise --clobber-config --config charImage.repair.doCosmicRay=True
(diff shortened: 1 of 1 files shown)


SimonKrughoff:
@jchiang87 `charImage.repair.doInterpolate` should definitely be turned on (set to `True`).  Most of the downstream code assumes the cosmic rays are not just detected but interpolated.

jchiang87:
Ok. I'll turn that on, but I note that when I ran `processEimage.py` with `--config charImage.repair.doCosmicRay=True` on the data for v222588-fu, the resulting S11.fits files were identical, with and without the `doInterpolate` flag set.  BTW, that's one of the visits where the the CR repair failed with `cosmicray.nCrPixelMax` left at its default value of 10000.

jchiang87:
Thanks, @SimonKrughoff .  @drphilmarshall Can you merge when ready please?

drphilmarshall:
Travis timed out again - not sure what we can do about that. Thanks Jim, thanks Simon!

jchiang87:
This closes #322, and closes #297 , as well.

Issue #309, opened 2016-09-08
Match Sprinkled AGN on magnitude


jbkalmbach:
Currently only matching source AGN on redshift. Need to add a requirement on magnitude as well.

drphilmarshall:
i-band mean AB magnitude to be exact (which OM10 is not).

jbkalmbach:
Hi @drphilmarshall, the columns in OM10 I'm wondering about are:

- MAGI_IN
- MAGI
- MAG (From previous discussion I know this is magnification, but just want to verify)
- ABMAG_I (I figure this is absolute and the one below is apparent, but magnitude of what?)
- APMAG_I

drphilmarshall:
Apologies for the nightmarish naming scheme - they were not my choices!

- MAGI_IN = source plane (ie unlensed, intrinsic) i-band AGN mean magnitude
- MAGI = i-band magnitude of 3rd brightest (in a quad) or 2nd brightest (in a double) AGN image
- MAG - magnification factors for each AGN image (ie MAGI = MAGI_IN - 2.5*log10(MAG[k]) where k is in the index of the 3rd (or 2nd) brightest image)
- ABMAG_I = Estimated absolute magnitude of the lens galaxy, in the (rest-frame) i-band
- APMAG_I = Estimated i-band apparent magnitude of the lens galaxy
- VELDISP = velocity dispersion of the lens galaxy, in km/s

I do not trust the ABMAG_I and APMAG_I: better to find a massive CatSim galaxy that has (approximately) the same redshift and velocity dispersion, and use that.

jbkalmbach:
Ok, so what I have put in place now after learning this is:
1) MAGI_IN values appear to be apparent magnitudes not the AB Mags you mentioned matching to above so I will match using apparent agn magnitudes as well.
2) Convert all om10 MAGI_IN values from i-band to imsimband to get mag_norm values.
3) Match sprinkled objects according to redshift +/- 0.05 and now additionally match agn_magnorm values to converted MAGI_IN values +/- 0.25 (what should this be?)

And then the final step will be adjusting the density parameter to account for the new matching requirements once we decide on a matching range for magnitude.

What do you think @danielsf @rbiswas4 @drphilmarshall?

jbkalmbach:
Code visible in branch: https://github.com/DarkEnergyScienceCollaboration/Twinkles/tree/issue/309/match_sprinkled_agn_on_magnitude

[...]

Issue #305, opened 2016-09-01
Write an instance catalog generation task


jchiang87:
@rbiswas4 This will entail adding the scripts you ran at NCSA to generate the instance catalogs (and SEDs) for Run1.1 to the Twinkles repository and packaging it all up as a task that we can run in batch via the SLAC workflow engine.   This will also entail updating the code to accommodate recent interface changes in the Stack, consistent with the version of the sims distribution we intend to use.

drphilmarshall:
I guess maybe @danielsf was working on this too - any news from up there? :-)

rbiswas4:
Yes. @danielsf has the branch working .... I plan to make some changes preserving that state, but to make life a little easier than before.

jchiang87:
closed by #327

Issue #313, opened 2016-09-15
Prepend instance catalog generation to phosim pipeline task


jchiang87:
Since we plan to generate the instance catalogs as a "pre-burner" to each phosim run, we need to modify the phosim pipeline to include that step.

TomGlanzman:
This is basically done but for one (hopefully final) problem.  

Request: a modification to generatePhosimInputs.py so that the "--seddir" string is NOT prepended to the file path stored in the instance catalog for each object, but continues to be used as the location in which to store the generated data products.

Reason:

The generatePhosimInputs.py script takes an option to specify the directory in which to store the SED files ("--seddir").  However, the script then prepends that directory + "spectra_files/" to each object's filename in the instance catalog.  By intention, I specify "--seddir" as an absolute path in a special semi-persistent store for pipeline efficiency.  When I run the script in the pipeline, objects in the instance catalog then have SED references that look like: /long-path/spectra_files/...

When phoSim runs, specifying the "--sed" option causes that string to be prepended to the object's SED reference in the instance catalog.  In the pipeline task, this is specified as an absolute path, e.g., /another-long-path.  The production SEDs have relative paths in the instance catalog, e.g., starSED/..., so this makes sense.  For the custom generated objects, that same treatment results in a bogus path, /another-long-path//long-path/spectra_files/...

jchiang87:
@TomGlanzman Can you make your request a separate issue and assign @rbiswas4 please?

TomGlanzman:
Done.  Issue #351

PR #327, opened 2016-09-29
Setups

### LINKED ISSUE #326: Is `TwinkesCatalogZPoint` different from `PhoSimCatalogZPoint` ?`
@jbkalmbach 

In  `Twinkles/python/desc/twinkles/generatePhosimInput.py` we use `Twinkles/python/desc/twinkles/twinklesCatalogDefs/TwinklesCatalogZPoint`. Is there a difference of this class with `PhoSimCatalogZPoint` in `sims_catUtils/python/lsst/sims/catUtils/exampleCatalogDefinitions/phoSimCatalogExamples.py`?

If this includes something more, should we be able to inherit most of the properties from `phoSimCatalogZPoint`?

rbiswas4:
-  added a method to further constrain matches on AGN systems in OM10 to CATSIM AGN using the magnitude of the source AGN in addition to the redshift (from @jbkalmbach ) #309 
- `twinklesCatalogDefs.py` updated to simplify definition of `TwinklesCatalogZPoint` and fix bug in #326
- Added code to in twinkles/twinkes_sky.py to (a) describe the change in Observation Metadata from OpSim as we change the number of snaps keeping the same exposure time. We also describe the astrophysical observables in `TwinklesSky` 
- An instance of this class takes as input OpSim pointings and locations to write out files, the tables (if need be) to specify what realization we are simulating. It has a method called  `writePhoSimCatalog` which writes out the phosim instance catalog corresponding to the object to a file. This is largely a refactoring of `generatePhoSimInput.py`  script modernized for the new catsim (@danielsf) which is still there in the twinkles directory. 
- setup scheme: added two files to declare Twinkles as an eups repository and setup twinkles. This enables scripts to use time_delay challenge data by looking in the Twinkles data directory. 
- Example script to use  `TwinklesSky`  to produce phoSim instance catalogs for Twinkles in 
bin/generatePhosimInput.py #305 

Partially addresses #305 #309 
Fixes #326 #329


setup/setup_locations_template.sh
```diff
@@ -0,0 +1,2 @@
+echo 'Copy this file to setup/setup_locations.sh and modify the directory to point to the directory of your OpSim output'
```
jchiang87: If we make the opsim db file a required argument of the `generatePhosimInput.py` script, this setup file, with its current contents, would not be needed.

bin/generatePhosimInput.py
```diff
@@ -0,0 +1,85 @@
...
+    obsHistID = obs_metaData._OpsimMetaData['obsHistID']
+
+    # all but first two are default values of optional parameters
+    # Kept in script to emphasize inputs
+    tSky = TwinklesSky(obs_metadata=obs_metaData,
+                       availableConnections=availConns,
+                       brightestStar_gmag_inCat=11.0,
+                       brightestGal_gmag_inCat=11.0,
+                       sntable='TwinkSN',
+                       sn_sedfile_prefix='spectra_files/specFile_')
```
jchiang87: Is this prefix relative to the cwd?  We might want some flexibility to allow the user to set the root path to which this is relative, i.e., via the `--seddir` option.

bin/generatePhosimInput.py
```diff
@@ -0,0 +1,85 @@
...
+import os
+from lsst.sims.catUtils.utils import ObservationMetaDataGenerator
+from desc.twinkles import TwinklesSky
+import argparse
+import pandas as pd
+from sqlalchemy import create_engine
+import time
+
+# Choose OpSim Output database
+OpSimDB = 'minion_1016_sqlite.db'
```
jchiang87: @TomGlanzman Can you have a look at and comment on this proposed interface for running it in the pre-burner stage discussed at #313 ?

python/desc/twinkles/twinkles_sky.py
```diff
@@ -0,0 +1,160 @@
...
+                                  TwinklesCatalogZPoint]
+
+        # SN 
+        ## SN catalogDBObject
+        if self.availableConnections is None:
+            self.availableConnections = list()
+            self.snObj = SNDBObj(table=sntable, connection=None)
+            self.availableConnections.append(self.snObj.connection)
+        else:
+            self.snObj = SNDBObj(table=sntable, connection=self.availableConnections[0])
```
danielsf: This is only valid if the `self.availableConnections[0]` is a connection to the database that `self.snObj` expects.  It will not be generalizable for users who want to use their own database (but maybe that is not something we want to support).

bin/generatePhosimInput.py
rbiswas4: Sure ... that should be easy to do.

python/desc/twinkles/twinkles_sky.py
```diff
@@ -0,0 +1,160 @@
...
+        t_after_starCat = time.time()
+
+        galCat = CompoundInstanceCatalog(self.compoundGalICList,
+                                         self.compoundGalDBList,
+                                         obs_metadata=self.obs_metadata,
+                                         constraint=self.brightestGalMag,
+                                         compoundDBclass=sprinklerCompound)
+
+        # galCat._active_connections = starCat._active_connections # pass along already open fatboy connections
+        galCat._active_connections = self.availableConnections
```
danielsf: The `CompoundInstanceCatalog` will open new connections if it has to (i.e. if it has to query a database that does not already have an open connection), by passing in `self.availableConnections` rather than `starCat._active_connections` you are not giving `galCat` access to all of the connections that `starCat` opened.  Again, this is not relevant to us, since all of our connections are to fatboy, but it might be relevant to a user who tries to use different databases for different data sets.

Feel free to ignore this comment if that is not a use case we want to support.

setup/setup_locations_template.sh
rbiswas4: Yes. This is correct(I was running it in that mode for a while but got tired of providing the argument everytime, and did this instead.)

bin/generatePhosimInput.py
rbiswas4: It is currently relative to cwd. Should make it absolute path or enable changing the root path?

python/desc/twinkles/twinkles_sky.py
rbiswas4: OK. I see what you are saying. 

BTW: what happens if there is only one active connection but when I append, does the list just keep increasing in size with copies of the same connection for every element?

bin/generatePhosimInput.py
jchiang87: I guess I would say that the code should prepend whatever the `--seddir` argument is to `sn_sedfile_prefix`.  So the default value of `--seddir` should be `.` (i.e., the cwd), and the user can override with either a relative or absolute path. Either way the line where `sn_sed_file_dir` is set would become
```
sn_sed_file_dir = os.path.join(args.seddir, 'spectra_files')
```

python/desc/twinkles/twinkles_sky.py
danielsf: I am suggesting that you use
```
galCat._active_connections = starCat._active_connections
```
`starCat` will only open a new connection if it needs it, so the list will not grow without bounds.

TomGlanzman:
@jchiang87 Your proposed interface looks fine.  How will I obtain the obsHistID?  A text file containing one obsHistID per line, for example, would be fine.

rbiswas4:
@TomGlanzman 
Perhaps,  you can start off with a csv file with OpSim records and use the line number to call this? I believe this will be most efficient. 
Also pinging @jchiang87 and @danielsf on this.

jchiang87:
We could also write a python function that provides a list of obsHistIDs from the opsim DB file itself based on the fieldID (i.e., fieldID=1427) corresponding to our Twinkles field.  I assume we aren't downselecting visits, so such a function would ensure consistency between the opsim db file and obsHistID requested.

jchiang87:
@TomGlanzman I put a copy of minion_1016.sqlite.db at SLAC here:
```
/nfs/farm/g/desc/u1/data/Twinkles/minion_1016_sqlite.db
```

python/desc/twinkles/twinkles_sky.py
rbiswas4: Done.

rbiswas4:
I think I have implemented all the requested changes from @jchiang87  and @danielsf.

About the conversation with @TomGlanzman and @jchiang87 about obsHistIDs:
We will downselect a bit (only to avoid duplicates in OpSim as I was doing previously)
Note, there is  a problem (not suggesting that this is worth worrying about, just want to make sure you are aware) that this would mean querying the OpSim for a record of each pointing which would take much more time than reading them off a file/memory. We could do the latter too.

bin/generatePhosimInput.py
```diff
@@ -0,0 +1,129 @@
...
+                        help='Visit number (obsHistID)')
+    parser.add_argument('--OpSimDBDir',
+                        help='absolute path to dir with the opsimBD',
+                        type=str,
+                        default='./')
+    parser.add_argument('--outfile', type=str, default=None,
+                        help='output filename for instance catalog')
+    parser.add_argument('--seddir',
+                        type=str,
+                        default=None,
```
jchiang87: I think this line would be better as
```
                       default='.',
```
so that line 120 below works in the default case. (I should have noted this change and the above in my example code when I made my last comment on the command line interface.)

bin/generatePhosimInput.py
```diff
@@ -0,0 +1,129 @@
...
+                        help='OpSim database sqlite filename',
+                        default='minion_1016_sqlite.db')
+    parser.add_argument('visit',
+                        type=int,
+                        help='Visit number (obsHistID)')
+    parser.add_argument('--OpSimDBDir',
+                        help='absolute path to dir with the opsimBD',
+                        type=str,
+                        default='./')
+    parser.add_argument('--outfile', type=str, default=None,
```
jchiang87: Since the default value of `None` will be passed as `fname` to `generateSinglePointing`, it would be good to add some code below to set `args.outfile` to something sensible.  I think one can do
```
    if args.outfile is None:
        args.outfile = 'phosim_input_%i.txt` % args.visit
```

bin/generatePhosimInput.py
```diff
@@ -0,0 +1,129 @@
...
+    sn_sed_file_dir = os.path.join(args.seddir, 'spectra_files')
+
+    availConns = None
+    print('will generate pointing for {0} and write to filename {1}'.format(
+          obs_metaData._OpsimMetaData['obsHistID'], args.outfile))
+    generateSinglePointing(obs_metaData,
+                           availableConns=availConns,
+                           sntable='TwinkSN',
+                           fname=args.outfile,
+                           sn_sed_file_dir=sn_sed_file_dir)
```
jchiang87: Does this work?  In the function declaration above none of the arguments to `generateSinglePointing` are keyword arguments.

bin/generatePhosimInput.py
rbiswas4: Sure ... I had a function doing that in the script, will put it back.

bin/generatePhosimInput.py
rbiswas4: OK ... I would agree :)

bin/generatePhosimInput.py
rbiswas4: I ran it (with the example code included in the description) and it worked. 

I don't understand the point you are making about keyword arguments. My expectations of the function declaration was that the function will fail if values are not provided for each parameter.

jchiang87: ah, I learned something!  I had thought it wasn't possible to treat non-keyword args as keyword args, i.e.,
```
def func(a, b):
    ...
```
could only be called like `func(my_a, my_b)`, but being able to do `func(my_a, b=my_b)` is very useful.

rbiswas4: I see. The reason I was confused is that I used to think that it was the entirely extensible dictionary of arguments that was called keyword arguments.

```
def func(a, b, **kwargs):
    ...
```
Some more googling explained to me that there is a usage in python where arguments that have names are also called keyword arguments.

python/desc/twinkles/generatePhosimInput.py
```diff
@@ -16,122 +16,166 @@
...
                                                  obs_metadata=obs_metadata,
                                                  # constraint='g_ab > 11.',
                                                  compoundDBclass=sprinklerCompound)
+
+                galCat._active_connections = starCat._active_connections # pass along already open fatboy connections
+
+                print("writing galCat")
+
                 galCat.write_catalog(filename, write_mode='a',
-                                     write_header=False, chunk_size=10000)
```
jchiang87: Revert back to this line so that `chunk_size=10000` is used for memory-limited machines.

jchiang87:
As a possibly unimportant aside, when I run `generatePhosimInputs.py`, there is a file created called `None` appearing in the working or output directory:
```
$ ls -lrt
total 178725
lrwxrwxrwx 1 jchiang desc       41 Oct  6 10:32 minion_1016_sqlite.db -> ../../data/Twinkles/minion_1016_sqlite.db
-rw-rw-r-- 1 jchiang desc 63423315 Oct  6 11:49 phosim_input_200.txt
-rw-rw-r-- 1 jchiang desc 68369720 Oct  6 12:47 phosim_input_220.txt
-rw-rw-r-- 1 jchiang desc 64047864 Oct  6 13:14 phosim_input_230.txt
-rw-rw-r-- 1 jchiang desc 64048055 Oct  6 13:54 phosim_input_250.txt
-rwxrwxr-x 1 jchiang desc      139 Oct  6 13:56 run_generatePhosimInput.sh*
drwxrwsr-x 2 jchiang desc     4010 Oct  6 14:09 spectra_files/
-rw-rw-r-- 1 jchiang desc    16703 Oct  6 14:09 None
-rw-rw-r-- 1 jchiang desc 64047994 Oct  6 14:09 phosim_input_276.txt
$
```
It looks like it's an SED file, though all of the y-values are zero.
```
$ wc None
 1771  3543 16703 None
$ head None
# Wavelength(nm)  Flambda(ergs/cm^s/s/nm)
30.00 0
31.00 0
32.00 0
33.00 0
34.00 0
35.00 0
36.00 0
37.00 0
38.00 0
$
```
and it seems to appear when the code reports
```
writing sne
```
to the screen.
Not a big deal, but I couldn't see it being created by `Twinkles` code anywhere.  If it is in some Twinkles code, we should put in a fix, but we shouldn't let this prevent this PR to be merged.

rbiswas4:
@jchiang87 Good catch. It is a catsim code for the sne which does this unintentionally. I do think it is harmless, but we should remove it (later) from catsim. This should be a lsst.sims ticket for me.

jchiang87:
BTW, after addressing the two remaining change requests that I made, I think this is ready to merge.

drphilmarshall:
Travis is failing, with
```
$ eups declare -r . twinkles -t current

eups declare: I'm unable to declare twinkles as tablefile
ups/twinkles.table does not exist
```

If this is not a problem, please do go ahead and merge.

danielsf:
This is a bug.  It should be `eups declare -r. Twinkles` as that is the name of the package and the table file.  We haven't caught it before because Mac's are not case-sensitive so `twinkles==Twinkles`.

danielsf:
At least, I suspect that is what is going on...

rbiswas4:
While that is a bug,  travis will probably fail from now on. We are trying to import functions that are in the latest `lsst.sims` installed at SLAC using `eups distrib install` but conda installs don't have this.

danielsf:
I guess another option would be to revert the table file names to lower case `twinkles`.  I changed them a few commits back, which may have been a mistake (I have been manually declaring a `Twinkles` package on my machine, which may have confused eups)

rbiswas4:
@jchiang87 I was checking that I am able to run on my computer. I noted that even though I made the script executable, it does not run on my El Capitan laptop using a 
`./script parameters` type command, but can be run using the 
` python script  parameters` command. I think this is a incompatibility of OSX SIP (El Capitan onwards) with the way the stack is set up (the same errors show up for trying to use the stack with `ipython`). Using the `python script arguments` methods, I checked that I can run the current script correctly.

Also @drphilmarshall, I do not have permissions to merge the PR (even though @jchiang87 approved it) You may have to do the merge (which both @jchiang87 and I am happy with).

drphilmarshall:
Thanks all!

PR #333, opened 2016-09-29
a version of segmentation.txt that has zero crosstalk

jchiang87:
In the `data/lsst` subdirectory of the phosim installation, the existing `segmentation.txt` should be moved out of the way and a symlink to or copy of `Twinkles/workflows/PhoSim/segmentation_3.5.3.txt` should replace it.  This file is intended for use with phosim 3.5.3 (or later).


SimonKrughoff:
I'm a little afraid that the formatting may break phosim parameter parsing.  I've found phosim to be fairly rigid with respect to parsing in the past.  I've made a file I'm using to test turning off crosstalk that adheres more closely to the formatting of the original.  Do you mind if we use that file instead, given that I can show that it actually serves to turn off crosstalk?

jchiang87:
I did run phosim 3.5.3 using this file.  It worked fine, but if you want to replace it with yours, I don't have any objection.

jchiang87:
Withdrawing this PR in light of comments at #330.

Issue #310, opened 2016-09-08
Sprinkle foreground lensing galaxies based upon match to Catsim mass and redshift


jbkalmbach:
Currently all lens galaxies in the sprinkler have the same radius and SED. We will improve upon this by finding CATSIM galaxies within an appropriate range of redshift and mass and taking the CATSIM size and SED to use as input for the sprinkled instance catalogs.

jbkalmbach:
From Scott: "You will need to be careful.  PhoSim (and thus CatSim) specify galaxy size as a semimajor and semiminor axis, so ellipticity is implied."

drphilmarshall:
If we have enough CatSim galaxies to select on ellipticity as well as mass and redshift that would be ideal. The OM10 catalog gives an ellipticity defined as `e = 1 - b/a` where `b` is the semi-minor and `a` the semi-major axis.

From the [CatSim schema](https://confluence.lsstcorp.org/display/SIM/Database+Schema) I'd say we want to match OM10 velocity dispersions based on `mass_stellar`. Is that a total stellar mass (eg associated with a Sersic model), or the mass within some aperture? 

The best way to match OM10 to CatSim is probably to get an empirical resampling in log10(mass_stellar) - vdisp - redshift from @astrosonnen (who has a model constrained by the SL2S and SLACS lens samples, in his [SL2S Paper V](http://arxiv.org/pdf/1410.1881v3.pdf)), and then work with that. That way we can replace the OM10 lens galaxy magnitudes (which I don't really trust) with CatSim galaxy magnitudes. Alessandro, can you help us out please? We just need a set of posterior samples drawn from Pr(logMstar, vdisp, zd | SL2S, SDSS) (ie marginalized over all your other parameters). Do you have this to hand, or could you make it for us quickly, please? Thanks!

rbiswas4:
> From the CatSim schema I'd say we want to match OM10 velocity dispersions based on mass_stellar. Is that a total stellar mass (eg associated with a Sersic model), or the mass within some aperture?

It should be total stellar mass associated with a galaxy (not sure if that is the same as what you mean by associated with a Sersic Model). 

Is there a reason why the velocity dispersion does not depend on the DM mass of the galaxy?

drphilmarshall:
It does, but it correlates more strongly with the stellar mass. We *could*
ask @astrosonnen for halo masses as well, but the existing strong lens data
do not constrain the relation between mass_halo and vdisp as well - so I
[...]

PR #334, opened 2016-10-03
Matching lens galaxies

drphilmarshall:
In response to issue #310. This PR is not yet ready to be merged, but it'll help me see what's changed. The main functional test is the [MatchingLensGalaxies.ipynb notebook.](https://github.com/DarkEnergyScienceCollaboration/Twinkles/blob/issue/310/match-lenses/examples/notebooks/MatchingLensGalaxies.ipynb)

This is looking pretty good so far, Bryce, although I thought we might need 2 or 3 Gaussians instead of just one. Could you please draw a few thousand samples from the GMM, plot them as a corner plot and overlay them with the datapoints, please? We need to make sure our model is generating sensible masses, redshifts and velocity dispersions.


jbkalmbach:
Ok, I've updated the notebook with a plot comparing 10000 samples with the test data. One thing that worries me is that some samples have the radius going below 0. Is there a way in empiriciSN to set a lower bound on a parameter?

drphilmarshall:
I think we can just throw out samples with radius less then some minimum
threshold, no problem. I am more concerned about the spread in vdisp not
being captured: what does the cornerplot look like if you choose 3
components?

jbkalmbach:
Hey Phil, not sure if you saw it or not, but I updated the notebooks with additional components in the GMM. It seems there is still some spread in vdisp not being represented by the training and I wonder if this may be a feature of breaking up a small dataset like ours?

drphilmarshall:
Thanks Bryce! Those extra components are crazy - I wonder why they are
being fitted like that? My first thought is to suspect the errors on the
data points: tiny errors on z and Reff could be driving the fit somehow.
You could try artificially inflating these error bars by a factor of 10, to
see if that made the 3-component GMM more stable.

It looks a bit like the two additional components are being "hidden", in
regions where they do not affect the likelihood evaluation much. This is
why they are disfavored by the BIC. Still, it's not clear to me why
additional components cannot be found that *do* improve the fit: perhaps
XDGMM is simply not exploring its parameter space well enough. Can you look
into initializing the 3-component model from a variety of different
starting positions (ie different, randomly selected, Gaussian component
means and covariances)? This might help us find better models.

Another possibility is that the small dataset is hurting us, by making
fitting difficult. There's not much we can do about that: a hack would be
to manually resample the data, by placing a 4D Gaussian of fixed width on
each data point, and sampling a few hundred fake datapoints from each one.
That would give us a big dataset that looked a bit like the input, could
probably be XDGMM-modeled more stably, but would be somewhat broader than
it should be - perhaps this is an approximation we could live with, though?
A straight astroML GMM with N_components = N_data might do this job very
quickly.

jbkalmbach:
I wrote up some code to match OM10 galaxies to CATSIM with our predicted stellar masses and found that we might want to go with the 3 component model after all. All my notes are in the notebook, take a look and let me know your thoughts.

drphilmarshall:
Interesting! Good that you can match most of the OM10 lens galaxies up with
a CatSim object with non-null bulge. I think it's OK to use either the 1 or
2 component model, but I have qualms about the 3 -component one: the
structure in the corner plot just doesn't look very plausible! If we were
to compromise and go with the 2-component model, and then argue in the
paper that all we need is a *plausible* population of lens galaxies, that
have the right colors and sizes, would we lose a lot of lenses? If its only
10% I think we can live with that. Great work Bryce - thank you!

jbkalmbach:
@drphilmarshall 2 components looks good. Just a couple questions on what we want to do now.

1) I plan on removing the OM10 lenses for which we don't get a match in CATSIM. Looks like it will only be 11, so that should be fine, but wanted to check. 

2) Then we want to get SED filenames and add those in with the OM10 data which seems like no problem. But when we get a_b and b_b values from CATSIM do we want to use those in Twinkles lens galaxies or do we want to use the a_b from CATSIM with the ellipticity from OM10?

3) Finally, I'm assuming we want to use the position angle from OM10 and not replace that from CATSIM, but just wanted to check.

drphilmarshall:
Great!

1) Yep, let's just drop the lenses we can't match.

2) I'd say use the CatSim size but the OM10 ellipticity. Best is to make
R_eff = \sqrt(a_b * b_b) and a = R_eff / sqrt(1- ellip_OM10) and b = R_eff
* sqrt(1- ellip_OM10)

3) Yes - definitely use the OM10 position angle.

Thanks Bryce! Nearly there :-)

jbkalmbach:
Ok, I updated the sprinkler to include the new sed filename and the a_b, b_b data from CATSIM. This is ready for review and merge @drphilmarshall.

jbkalmbach:
Oh, also, just to note. By trimming the catalog to remove the ones without CATSIM matches we lost one system when sprinkling. So we currently have 198.

examples/notebooks/MatchingLensGalaxies_utilities.py
```diff
@@ -0,0 +1,249 @@
...
+
+    plt.xticks(param_range,fontsize=14)
+    plt.yticks(fontsize=14)
+
+    plt.xlabel('Number of components',fontsize=18)
+    plt.ylabel('BIC score',fontsize=18)
+
+    plt.show()
+
+def get_demo_data(flist):
```
drphilmarshall: This def and the one below it look like vestigial code. Can they be excised?

data/SonnenfeldEtal2013_Table3.csv
```diff
@@ -0,0 +1,26 @@
+#ID, zlens, vdisp, Reff, Mstar, zlens_err, vdisp_err, Reff_err, Mstar_err
+SL2SJ021247-055552, 0.750, 267, 8.92, 11.45, 0.001, 17, 0.01, 0.17
```
drphilmarshall: Reff of 8.92 looks like an effective radius in kpc, not arcsec. Can you please investigate, and then correct the code below if necessary?

python/desc/twinkles/sprinkler.py
```diff
@@ -175,14 +175,13 @@ def sprinkle(self):
...
-                    row['galaxyBulge_sedFilename'] = self.LRG_name
+                    row['galaxyBulge_sedFilename'] = newlens['lens_sed']
                     row['galaxyBulge_redshift'] = newlens['ZLENS']
                     row['galaxyDisk_redshift'] = newlens['ZLENS']
                     row['galaxyAgn_redshift'] = newlens['ZLENS']
                     row['galaxyBulge_magNorm'] = matchBase().calcMagNorm([newlens['APMAG_I']], self.LRG, self.bandpassDict) #Changed from i band to imsimband
-                    newlens['REFF'] = 1.0 #Hard coded for now. See issue in OM10 github.
-                    row['galaxyBulge_majorAxis'] = radiansFromArcsec(newlens['REFF'])
-                    row['galaxyBulge_minorAxis'] = radiansFromArcsec(newlens['REFF'] * (1 - newlens['ELLIP']))
+                    row['galaxyBulge_majorAxis'] = radiansFromArcsec(newlens['REFF'] / np.sqrt(1 - newlens['ELLIP']))
```
drphilmarshall: We may need to convert Reff in kpc into arcsec first (see previous comment)...

doc/LSST_DESC_Notes/0001_twinkles_design/index.md
```diff
@@ -1,104 +1,86 @@
 ![](./_static/header.png)
 
-LSST DESC Notes Template and Author Guidelines
-==============================================
+# Twinkles: Science Goals, Survey Design and Development Methodology
```
drphilmarshall: These design note edits look like mine, from the design-note branch... Some pollution seems to have occurred. It looks harmless though.

python/desc/twinkles/sprinkler.py
jbkalmbach: Concerning the Reff. I decided to use the values from CATSIM for the galaxies we matched to and not use these values. When looking at the notebook predictions for Reff in the test set the errors were very large compared to the errors we found in stellar mass (see cell 11 and 12 in the notebook).

drphilmarshall: Ah! OK good :-) What do the typical CatSim galaxy sizes (in arcsec) turn out to be, though? I'm expecting ~1 arcsec or so.

jbkalmbach: r_eff

Above is a plot of the REFF column in the new Twinkles lens catalog. The median is at .73 so it looks like they skew smaller than 1 arcsec.

drphilmarshall: Great! Looks good to me. Thanks for the plot, Bryce :-)

drphilmarshall:
OK, this looks good to go to me. Weirdly I don't seem to be able to "approve" the changes in the review, despite being an admin. I'll merge it anyway.

drphilmarshall:
Ohhhh it's because I *opened* the PR! Hah. I guess that is fair enough: as a GitHub PM you can either be impatient, or in control, but not both :-)

PR #335, opened 2016-10-04
Tools to extract obsHistIDs from the OpSim db file.

jchiang87:
This includes a module `get_twinkles_visits` and a command line tool `get_twinkles_visits.py` and addresses https://github.com/DarkEnergyScienceCollaboration/Twinkles/pull/327#issuecomment-250844496 and subsequent discussion.


python/desc/twinkles/sqlite_tools.py
```diff
@@ -29,3 +31,30 @@ def create(self, columns, table, condition=None):
             query += ' ' + condition
         data = [row for row in self.conn.execute(query)]
         return pd.DataFrame(data=data, columns=columns)
+
```
rbiswas4: Is there a reason to use this rather than use `pd.read_sql_query` or `pd.read_sql_table` ?

jchiang87: Not particularly.  I'm just not very familiar with the available pandas functions.  I guess in this case, `pd.read_sql_query` would be the one to use assuming we really didn't want to ingest the entire table?

rbiswas4: @jchiang87 One different thing I was worried about:  Should we try to drop duplicates, I think this query could pull in two records for the same pointing with two different propIDs. (This is taken care of in the script with a `DISTINCT` in the query, but I don't see it in the dataFrames). Also if `propID` is  becomes a requested column, a simple `DISTINCT` will not do as the propIDs are different.

jchiang87: This tool was meant to be a sort of general purpose way of packaging columns from an sqlite file into a data frame.   I'll add access to the `DISTINCT` flag, but more generally, once the user has the data frame, any subsequent selection, including down-selecting rows, can be easily done using the pandas selection facilities.  I guess one could argue that this class is trivial enough that it really is not needed.

rbiswas4: @jchiang87 I understand that, but am worried that someone will try to use this to get dataframes from the opsim database and not notice the duplicates. In principle, since the fact that these duplicates exist are in the OpSim documentation, you could say it is the user's responsibility ...

jchiang87: At some point, the user does need to understand the content of the data they are using, so since the fact that the duplicates exist is documented, I think that's sufficient.

TomGlanzman:
Is there now a favored tool for providing a list of obsHistIDs or is this still under discussion/development?

jchiang87:
There's a tool here:  #335 Feel free to comment on the interface.

rbiswas4:
@jchiang87 @TomGlanzman @drphilmarshall @jbkalmbach 

One thought that popped up in my head was that we would probably like to order the obshistids in the Twinkles Run3 in a way that if we have the first 1000 of the 25000 done, we can start doing some early analysis. 

For example for SN light curves, it might mean getting a decent set of the first (replace by nth year) year light curves first (la Twinkles 1). And then filling up the first year light curves. Are there similar thoughts for strong lensing analyses? Would it make sense to have this in Twinkles 3? I will try adding this to today's agenda.

PR #336, opened 2016-10-06
Finalize sprinkler parameters for AGN

jbkalmbach:
Use final numbers agreed upon in #309


python/desc/twinkles/sprinkler.py
```diff
@@ -181,8 +181,8 @@ def sprinkle(self):
 
     def find_lens_candidates(self, galz, gal_mag):
         # search the OM10 catalog for all sources +- 0.05 in redshift from the catsim source
```
drphilmarshall: Selection looks good, but comment needs updating to match!

jbkalmbach:
I'll merge this in an hour if there are no more comments.

drphilmarshall:
Thanks Bryce! :-)

Issue #312, opened 2016-09-08
Avoiding phoSim indexing failures in Run 3


rbiswas4:
@drphilmarshall 
Is there a plan to avoid indexing failures due to phosim for run3 ? As a reminder in previous runs, we have found and reported that PhoSim tries to index objects by floats, and indices are messed up due to precision problems when these are large. I do not know if phosim was changed to avoid this problem. I forgot to mention this during the meeting. 

A possibility could be checking out our own branch of phosim and modfiying the index to a string before writing it out. We can then tag this and use this tagged version. I will not have time to work on it immediately.

drphilmarshall:
Hmm. Do you know if phosim has been upgraded to use string or integer
indices, @TomGlanzman? Does anyone know whether there is an open ticket on
the phosim release bitbucket?

jchiang87:
This doesn't seem to have been issued there:
https://bitbucket.org/phosim/phosim_release/issues

drphilmarshall:
Good, thanks Jim. Rahul or Bryce, can you please write a phosim ticket and
ask for advice regarding a workaround? When you've made it, please post the
link to it here so we can follow along. Thanks!

rbiswas4:
I thought I saw that Bryce had already posted a link but : https://jira.lsstcorp.org/secure/RapidBoard.jspa?rapidView=12&projectKey=PHOSIM&view=detail&selectedIssue=PHOSIM-27

rbiswas4:
Closed through #340 for run 3 and kept track of in genral in #383

PR #340, opened 2016-10-07
U/danielsf/mangle galtileid

danielsf:
this PR should solve the problem of AGN object IDs getting misrepresented in PhoSim

### DIFF
--- a/python/desc/twinkles/sprinkler.py
+++ b/python/desc/twinkles/sprinkler.py
@@ -28,6 +28,17 @@ class sprinklerCompound(GalaxyTileCompoundObj):
             if 'raJ2000' in name or 'decJ2000' in name:
                 results[name] = np.radians(results[name])
 
+        # the stored procedure on fatboy that queries the galaxies
+        # constructs galtileid by taking
+        #
+        # tileid*10^8 + galid
+        #
+        # this causes galtileid to be so large that the uniqueIDs in the
+        # Twinkles InstanceCatalogs are too large for PhoSim to handle.
+        # Since Twinkles is only focused on one tile on the sky, we will remove
+        # the factor of 10^8, making the uniqueIDs a more manageable size
+        results['galtileid'] = results['galtileid']%100000000
+
         #Use Sprinkler now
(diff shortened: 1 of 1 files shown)


jbkalmbach:
Did you do a check to confirm unique Ids in the Twinkles area?

danielsf:
I generated one InstanceCatalog and verified that

1) the uniqueIDs were all less than 10^16
2) the IDs were unique

but just for one InstanceCatalog

danielsf:
Actually... I tested "a pointing".  It wasn't actually the Twinkles field

danielsf:
I am generating a Twinkles field now

danielsf:
The actual Twinkles field also passes the test.  The largest IDs are now the supernova IDs, which, by definition, can't go higher than 7*10^15

jbkalmbach:
Looks good. I've confirmed the Lensed AGN systems id numbers still contain all the correct information and there are the same number of them.

danielsf:
@jbkalmbach It looks like I am not authorized to push to this repo.  You will have to merge the PR for me.

jbkalmbach:
@danielsf Looks like I can't either actually. @jchiang87 @drphilmarshall Can one of you help us?

jchiang87:
I don't have authorization either.   @drphilmarshall help!

danielsf:
At least we don't have to worry about accidentally force-pushing to master....

drphilmarshall:
Wow. I feel like I have too much power, here. I'll take a look at the settings. Thanks for the bugfix, Scott! :-)

drphilmarshall:
OK, FYI: I removed the restriction on who can push to master (it was set to be only the DESC admins, by default), and replaced it with the requirement that a review "Approval" must be given instead. I think this makes more sense for us.

rbiswas4:
This also closes: https://github.com/DarkEnergyScienceCollaboration/Twinkles/issues/312 which is specific to run 3, since we are not going to be able to do much else. However, we should keep track of this issue for a more graceful solution in future maybe.

Issue #346, opened 2016-10-13
PhoSim instance catalogs: speed up galaxies with pre-cooked FatBoy output


drphilmarshall:
@danielsf You and @rbiswas4 had this good idea, of speeding up the PhoSim instance catalog generation by not asking FatBoy to repeatedly go find the Twinkles galaxies every visit. Here's an issue to track your work on implementing this! :-)

@cwwalter We noted in the Twinkles meeting that this approach should also speed up the instance catalog generation for DC1 PhoSim deep as well, which is cool. We are finding that each IC generation takes about 20 mins right now, which is a significant (if not dominant) expense.

rbiswas4:
> by not asking FatBoy to repeatedly go find the Twinkles galaxies every visit.
Actually the idea is not to talk to fatboy at all. 

Also @danielsf, if you do go on to think about doing tiling on the user side, do let me know, I probably have most of that code before I decided to do other stuff first.

danielsf:
I am almost ready to submit a pull request addressing this issue.

The cache of Twinkles galaxies comes out to about 80 MB as a text file.  Are we okay just making that  a part of the repo, or should I try to do something clever with our setup scripts so that users generate the galaxy cache whenever they install Twinkles?

jchiang87:
Can you add a script to generate the file so that the users have the option do it themselves?  And if it doesn't exist, can you have the code access fatboy in the usual way?  If that's a lot of work, then I'd say just commit the file.

danielsf:
How about: if the file does not exist, I will call the code to generate it?  That should be pretty straightforward.

jchiang87:
ok, sounds good to me.

rbiswas4:
Slightly unrelated (since the suggested solution works well for the present case). We have and will be facing a more generic version of this problem often. Our examples and integration tests will depend on input files that are at least as large if not larger (1GB +) once we start adding in DM analysis.  This will likely hit us during hack week. Should we plan to allocate some data space on a web server?

PR #349, opened 2016-10-18
U/danielsf/cache galaxies

danielsf:
This should address issue #346 

When users run generatePhoSimInput.py, the code checks for a sqlite database $TWINKLES_DIR/data/galaxy_cache.db, containing all of the galaxies in the Twinkles field of view.  If this database is not there, the code creates it.  This local database is then used to generate the galaxies in the PhoSim InstanceCatalogs.


jbkalmbach:
This looks good to me. I'm not going to request that this be done before merging this, but I'm making a new issue where we update the repo [setup instructions](https://github.com/DarkEnergyScienceCollaboration/Twinkles/blob/master/doc/Setup.md) because I didn't notice before, but it looks like we have removed "generatePhosimInput.py" from our pipeline and moved it into twinkles_sky?

jchiang87:
@rbiswas4 @jbkalmbach Are you guys ok with these changes?  If so, I think we should go ahead and merge.

jbkalmbach:
@jchiang87 @danielsf Yeah unless Rahul has any objections let's merge it.

rbiswas4:
Only one comment: 
Are we changing the units in which the ra, dec are stored obtained ? It used to be in radians
but the class GalaxyTileObjDegrees(GalaxyTileObj) suggests that it is in degrees?  More generally, it might be good to add a docstring with the units. 

Otherwise this is good to merge.

danielsf:
CatSim expects ra, dec to be stored externally (i.e. on fatboy) in degrees.  Because GalaxyTileObjDegrees exists solely to produce a stand-in for fatboy, it needs to output ra, dec in degrees.

rbiswas4:
Ah ... so the classes will work in the same way. OK got it!

PR #353, opened 2016-10-19
Issue#352/add cat sim instance catalog reading to io

### LINKED ISSUE #352: Add io tools to read catsim Instance Catalogs into a `pd.DataFrame` correctly
CatSim Instance Catalogs can only be written to disk in the form of csv files. However, the header of the csv file has both commas and spaces, along with a `#` sign at the beginning and is not the standard format  of csv files that are read into `pandas.dataframe`. Add a io module where a function is provided to do this easily

rbiswas4:
Quick PR: Have been using this in too many branches and want to have a function to go from Instance Catalogs to dataframes. Closes #352


python/desc/twinkles/io.py
```diff
@@ -0,0 +1,25 @@
+from __future__ import absolute_import, print_function, division
```
jchiang87: Although this code is imported into the `desc.twinkles` namespace, it might be a good idea to rename the file to something less generic.  There already is an `io.py` module that's part of the standard python2 distribution.

rbiswas4: Would `twinkles_io` be ok? I agree that io.py is probably too generic, though it does not matter a lot, and I would be happy to change it to that or any other suggestions.

jchiang87: Sure that would be fine.

PR #356, opened 2016-10-19
Avoid lens CATSIM galaxies when adding SNe

jbkalmbach:
This pull request adds a data file with the galtileid info of lens galaxies used in the Twinkles field. This way we can tell CATSIM which galaxies to avoid when adding supernovae to the Twinkles Instance Catalogs.


drphilmarshall:
Bravo! Glad to see the back of this little bugbear :-) Thanks both.

PR #357, opened 2016-10-19
Added info on centroid files to lessons learned

### LINKED ISSUE #168: Centroid file SourceID numbers for lensed AGN differ from Phosim Input Catalog id numbers
Quoting from #155:

> @rbiswas4 and I took a look at this and it seems that it might be caused by a precision error when phosim converts all ids to a float. We looked at a small example phosim output and found that the id of an object was off by a small integer number of values in the centroid file from the phosim catalog. Rahul, I think you mentioned that you saw this before?

Is this in fact the problem? And if so, what can we do about it?

jbkalmbach:
This is to close #168.


doc/lessons_learned.md
```diff
@@ -22,3 +22,4 @@
 * **At NERSC, it matters a lot for throughput where the software is installed.** (Tony)
 * **We need to stay on top of the DMstack (apps and sims) development** both in terms of determining the versions to install but also to sort out problems as they are encountered. This is best done either through [community.lsst.org](http://community.lsst.org) or via GitHub issues (Heather)
 * **Conda installs of DMstack are great, but may not aways be available** In that case, brush up on your use of [lsstsw](https://developer.lsst.io/build-ci/lsstsw.html). (Heather)
+* **PhoSim Centroid files list sourceID incorrectly for large integers** It seems that larger id numbers go outside the default integer length in PhoSim and don't get stored in PhoSim Centroid files correctly. We solved this in Twinkles by making all our id numbers smaller by 10^8 because they can still be unique in our smaller field, but this is still an [open issue in phosim](https://jira.lsstcorp.org/browse/PHOSIM-27).
```
danielsf: I would make it more explicit, something like "In PhoSim, id's are stored as double precision floats.  This means that any number greater than 2^53 ~ 9*10^15 gets rounded off."  That way, we know what the limitations are.

Other than that, go ahead and merge.

Issue #72, opened 2015-12-11
Implement part of the cookbook as a Continuous Integration test


jchiang87:
Presumably, the steps from `ingestImages.py...` onwards in the [cookbook](https://github.com/DarkEnergyScienceCollaboration/Twinkles/blob/master/code/twinkles_cookbook.md) (and adding the running of `plot_point_mags.py`) could be crafted in to a single executable and run as an integration test in Jenkins.

Lower level unit tests would also be good to have.

SimonKrughoff:
I think we could even skip ingestImages.py and just start with a minimal input schema similar to the one I tarred up for the cookbook.

drphilmarshall:
@jchiang87 Do we want to resurrect this issue since we'll be working with a stack installed from source at SLAC? You had mentioned wanting to look into Jenkins again.

jchiang87:
I have Jenkins configured to do builds of our repos, and Brian set up a Slack channel for those builds: https://lsstc.slack.com/messages/desc-builds/

I've set up the webhook for pserv, but I need the admins of Twinkle and Monitor to do the same for those repos. @drphilmarshall @jbkalmbach  Could you set those up please?

To do that do
```
Settings -> Integrations & services -> Add service -> Jenkins (GitHub plugin)
```
and then for the Jenkins hook url, enter
```
http://srs.slac.stanford.edu/hudson/github-webhook/
```

jchiang87:
If someone really wants to do this they can either reopen or make a new issue.

PR #359, opened 2016-10-20
Issue#347/ordering obs hist i ds for twinkles3

### LINKED ISSUE #347: Ordering obsHistIDs For Twinkles 3
At the weekly meeting today we discussed a plan for ordering the obsHistIDs. The rationale behind this plan is that Twinkles 3 has ~ 25000 visits, which will probably take an appreciable amount of time, and we would like to be able to run meaningful analyses at intermediate times. So, the idea is to order the obshistids such that at intermediate times much shorter than the total walltime, we obtain sets of images which enable some analyses (in supernova and strong lensing analyses).  So the factors for ordering are
- Science analyses requiredments from SN and SL (SL has a preference for WFD, SN has a preference for complete light curves)
- PhoSim run times estimated via @sethdigel 's phosim time predictor: the idea being that we will try out the quicker runs first in the normal queues and think about other runs later, as our technology improves. 

This is the order that we provisionally agreed on:
- Drop duplicates: choose propid for WFD when duplicates exist
- We will select every unique band and night combination for run 3.1 and order them by expMJD first and then band. 
- In the case of multiple visits satisfying this combination of unique night + band:
   we select a WFD visit if possible, unless the phosim run time estimate exceeds a threshold T1
   If no WFD visit is selected select DDF visit with lowest phosim runtime.

My proposed choice of T1 is 120 Hours. Comments welcome!

rbiswas4:
This is a PR to address #347 

The PR includes code to do the Twinkles splitting, an example notebook, tests and a script. This is not yet ready to merge and the following changes are needed and I would like to get some opinions/help on these:

- Do we want all the obshistIDs to be written to one file, or the three parts separately ( @TomGlanzman @jchiang87 )
- Currently the class `OpSimOrdering` has a line which assigns random numbers to the columns `predictedPhoSimTimes` . Once I have @sethdigel pickle file, I will replace this column with predictions from his code.
- Tests: the test requires a path to an opsimdb. Right now it is hard coded for my computer. Is there a suggestion on how to do this? @jchiang87 

When done this will fix #347


TomGlanzman:
Hi @rbiswas4,  a single file is what I will eventually need, but concatenating 3 files would also be acceptable.  If a single file, feel free to insert "# Begin section 1" sorts of comments, as needed.

Will this code be incorporated into the obsHistID extractor or applied against its output?

rbiswas4:
Currently, the obsHistID extractor `bin/get_twinkles_visits.py` has this code built in to create a single file. I have not put the #Begin sections though, so, I will work on that.

rbiswas4:
I think this is ready for review. 
- Suggestions on what plots you would like to see in the example notebook (examples/notebooks/Example_OpSimOrdering.ipynb)  of Observation characteristics of visits in Twinkles 3.1 welcome (@sethdigel, @drphilmarshall , @jbkalmbach , @ksk)

sethdigel:
Very nice, @rbiswas4!  If it is easy, in the notebook I'd recommend also tabulating the numbers of 'too long' visits by filter.  For the notebook, to illustrate the effects from re-ordering it would be nifty to see a plots of numbers of runs completed vs. cumulative CPU time, based on the estimated CPU times, comparing two cases - one with the visits run in chronological order and one with the visits considered in their re-ordered versions.  The visits will not be run sequentially, but with many going in parallel, but still the plot would show how much (estimated) CPU time will be required to get to any desired level of completeness.

rbiswas4:
@sethdigel
tabulating the numbers of too long visits with filters is easy. 

I was thinking about making the other plot your refer to as well, but balked on two thoughts:
- The parallel vs serial issue you mentioned
- The re-ordering does not (necessarily) give us the best number of runs completed with cumulative CPU time, but is designed to give us a good coverage in terms of times and filters and was not sure how to do it. Trying to write this reply, I realize that the correct comparison to bring this out would be the number of 'unique' combinations of night and filters completed in the chronological order. So I now, know what I should do.

tests/test_opSimOrdering.py
```diff
@@ -0,0 +1,76 @@
...
+import numpy as np
+import pandas as pd
+from desc.twinkles import OpSimOrdering
+from lsst.utils import getPackageDir
+import unittest
+
+class TestOpSimOrdering(unittest.TestCase):
+
+    def setUp(self):
+        self.opSimDBPath = '/Users/rbiswas/data/LSST/OpSimData/minion_1016_sqlite.db'
```
jchiang87: Maybe we should have this be accessed from an environment variable.   I would really like to get this package running regularly in Jenkins at SLAC, but I need someone with admin privileges to set up the web hook as I described [here](https://github.com/DarkEnergyScienceCollaboration/Twinkles/issues/72#issuecomment-253580705) (@drphilmarshall Could you find a moment to do that please?).

tests/test_opSimOrdering.py
```diff
@@ -0,0 +1,76 @@
+import os
```
jchiang87: Thanks for adding unit tests for this work!  I'm a little concerned that this test code is essentially doing the same full processing of the opsimdb file that the `get_twinkles_visits.py` executable would do.  When I ran the executable just now, it took 20 mins on a generic rhel6-64 box at slac from the interactive pool, which seems too long for a set of unit tests.  (The unit tests are still cranking away after 12mins).  For testing purposes, I think it would be preferable to have a small opsimdb file that's a tiny subset of the full minion_1016 file.  We could put it in the `data` folder and so not have to worry about setting `self.obsSimDBPath` in a context-dependent way.

tests/test_opSimOrdering.py
rbiswas4: Yes, I think this can be done by reverting to the setup file that I had created in the generate PhoSim Input script pull request..

tests/test_opSimOrdering.py
rbiswas4: True. I can speed it up by using a `setUpClass`, but I will have trouble doing it this week since I am away. Another speedup is using a small OpSim output which I use for my testing in other repos. The small OpSim output I use is [here](https://github.com/rbiswas4/OpSimSummary/blob/master/opsimsummary/example_data/enigma_1189_micro.db) and is created by the [script](https://github.com/rbiswas4/OpSimSummary/blob/master/scripts/make_smallOpSim.py)

bin/get_twinkles_visits.py
```diff
@@ -4,20 +4,57 @@
...
+from lsst.utils import getPackageDir
 
 parser = argparse.ArgumentParser(description="Write an ascii file of visits derived from an OpSim db file given a fieldID")
 parser.add_argument('opsimDB', help='OpSim database sqlite file')
 parser.add_argument('--fieldID', type=int, default=1427,
                     help='ID number of the desired field')
 parser.add_argument('--outfile', type=str, default=None, help='output file')
+parser.add_argument('--orderObsHistIDsByDesign', type=bool, default=True,
+                    help='order obsHistIDs as decided for Twinkles 3')
+parser.add_argument('--randomForestPickleFileName', type=str, default='RF_pickle.p',
```
rbiswas4: I am confused. Are you saying provide an argument for the absolute path? (I can do that). or are you suggesting changing the variable names (can do that too).

rbiswas4: Also, I originally added the `ObsHistID.csv` to get a list of obsHistIds to Tom (and everyone else),  but of course, he is welcome to run it and generate it himself.

rbiswas4: And I agree about it being run 3 specific. I don't know how to refactor it more generally

tests/test_opSimOrdering.py
jchiang87: Yes, I think just using the small OpSim file is the way to go here.  Even doing `setUpClass` really won't be enough of a speed up.

tests/test_opSimOrdering.py
jchiang87: If we use a tiny opsim file as we discussed in the other comment, then set the path by doing
```
self.opSimDBPath = os.path.join(getPackageDir('Twinkles'), 'data', <small opsimdb filename>)
```
we can avoid the extra setup step.

bin/get_twinkles_visits.py
jchiang87: So I actually have two suggestion here, neither of which are compulsory:
* Just have a single option for the random forest file location, i.e., `--RF_pickle_file`.  This could be set to `os.path.join(getPackageDir('Twinkles'), 'data', 'RF_pickle.p')` if the default is `None`.
* make the option names shorter and easier to type.

TomGlanzman:
@rbiswas4 and @jchiang87 an operational question about get_twinkles_visits.py:  the script wants the RF_pickle.p file.  Jim suggested it would be in Twinkles/data but I do not find it in 'master', even after this merge.  The question is how should this command be run for Twinkles run3?  So far, I have this for running at SLAC:

.../Twinkles/bin/get_twinkles_visits.py  /nfs/farm/g/desc/u1/data/Twinkles/minion_1016_sqlite.db

and the script complains about RF_pickle.p's absence.  If not needed, how should the command be executed?

jchiang87:
There's a copy here:

/nfs/farm/g/lsst/u/digel/Twinkles/python/desc/twinkles/RF_pickle.p

Since this file is being used by the unit tests and its absence is breaking them, I guess I would add it to the data folder in the repo.  Can you do that in your branch and push to github?

TomGlanzman:
Thanks @jchiang87 , have added to my branch.  Also, the script now runs to completion.  However, with default options, it yields only 5784 total visits.  One such option is --maxPredTime=100.0 (hours).  Am running again with --maxPredTime=1000.0 just to see if all ~25,000 visits show up.  If so, then the 100.0 hours seems a harsh cut.  (At SLAC we could safely set the value to 120 hours to remain within the batch queue limit.)

Okay, the 2nd test ran for 20 minutes and produced a file with 6556 visits (with 1000 hour limit).  This may be a question to the group: are we happy with a ~6000 visit run3?  @drphilmarshall ?

TomGlanzman:
Ran another test with --maxPredTime=10000.0 (hours) and got the same 6556 visit result as with 1000.0 hrs.  Am I using the correct obssim database?  (/nfs/farm/g/desc/u1/data/Twinkles/minion_1016_sqlite.db)

rbiswas4:
The numbers I originally had (with the 100 hr limit) were 1508, 2138, 18324 with (2109 undone because of the wall clock limit). A place to see this is https://github.com/DarkEnergyScienceCollaboration/Twinkles/blob/master/examples/notebooks/Example_OpSimOrdering.ipynb. 

Bug found and fixed in #385

sethdigel:
Yes, the numbers in Rahul's notebook looked consistent with what we saw in Twinkles Run 1, for which about 10% of the visits could not be completed because of the 120-hour CPU time limit.  I had suggested that Rahul use 100 hours as the predicted CPU time limit because the scatter was somewhat greater with phosim 3.5, and because the down side of guessing wrong about the CPU time limit was that a lot of CPU time went for nothing.  And I thought that the  point of the sorting was that the visits were just being reordered for running on the batch farm, not that some were not going to be attempted.

Issue #361, opened 2016-10-20
Can we remove `https://github.com/DarkEnergyScienceCollaboration/Twinkles/blob/master/python/desc/twinkles/generatePhosimInput.py`


rbiswas4:
This seems to be outdated in favor of `bin/generatePhosimInput.py`. Do people think this can be deleted to avoid confusion?

jchiang87:
yes, definitely. There are a handful of other things I'd like to clean up regarding out-of-date code.

jbkalmbach:
Yes, this seems like a good idea.

jchiang87:
I'll make a PR from a branch where I've removed stuff that looks outdated.

drphilmarshall:
Great, thanks Jim!

PR #365, opened 2016-10-20
Clean up old code and standardize imports

jchiang87:
Also suppress `UserWarning`s from sims_catalogs code.

This addresses #361. 

@rbiswas4 I deleted `python/desc/twinkles/InstcatGenerationBooKeeping.py` since `python/desc/twinkles/validate.py` seems to supersede it, but I don't see `validate.py` being used anywhere.  Do you have plans for it?

@sethdigel I modified the way `python/desc/twinkles/phosim_cpu_pred.py` imports from `sqlite_tools` to obey the same conventions as the other modules in that directory.

@danielsf The warnings filter I put in `python/desc/twinkles/__init__.py` suppresses warnings like
```
/Users/rbiswas/src/LSST_el_Capitan/sims_catalogs/python/lsst/sims/catalogs/db/dbConnection.py:439: UserWarning: Duplicate object type id 25 specified: 
Output object ids may not be unique.
This may not be a problem if you do not want globally unique id values
  'want globally unique id values')
```
There are 9 or so of them currently.  Is there a better way of dealing with them so that they don't appear?  Or should we not be ignoring them?


rbiswas4:
@jchiang87 deleting the InstCatBookKeeping is fine. I had it when I generated files for Twinkles 1, we are not using the same setup, so I don't think it is useful any more.

As for validate, I do have plans for that, based on the reader for instancecatalogs, which allows us to compare values. That is on once I finish @TomGlanzman  (which ran into some trouble but should be done)  and the ordering. If @sethdigel 's prediction model API changes that would be useful to know before adding to the ordering problem

PR #366, opened 2016-10-21
Issue#351/spectra file name defs

### LINKED ISSUE #351: change to generatePhosimInputs.py to put only relative paths in instance catalogs
Request: 

A modification to generatePhosimInputs.py so that the "--seddir" string is NOT prepended to the file path stored in the instance catalog for each object, but continues to be used as the location in which to store the generated data products.

Reason:

The generatePhosimInputs.py script takes an option to specify the directory in which to store the SED files ("--seddir"). However, the script then prepends that directory + "spectra_files/" to each object's filename in the instance catalog. By intention, I specify "--seddir" as an absolute path in a special semi-persistent store for pipeline efficiency. When I run the script in the pipeline, objects in the instance catalog then have SED references that look like: /long-path/spectra_files/...

When phoSim runs, specifying the "--sed" option causes that string to be prepended to the object's SED reference in the instance catalog. In the pipeline task, this is specified as an absolute path, e.g., /another-long-path. The production SEDs have relative paths in the instance catalog, e.g., starSED/..., so this makes sense. For the custom generated objects, that same treatment results in a bogus path, /another-long-path//long-path/spectra_files/...

rbiswas4:
The PhoSim Instance Catalogs outputs a string to the path of the spectra files relative to `seddir` argument supplied to the `bin/generatePhoSimInstanceCatalog.py`  rather than the absolute path. 
 
`TwinklesSky` now uses `TwinklesPhoSimCatalogSN` rather than `PhoSimCatalogSN` directly. 
`TwinklesPhoSimCatalogSN` inherits from `PhoSimCatalogSN`, but 
- defines a getter for `shorterFileNames` which is `sedFilepath` with prepended seddir stripped as requested.
- redefine `column_outputs` to include `shorterFileNames` rather than `sedFilepath`

 

Fixes #351

### DIFF
--- a/python/desc/twinkles/twinklesCatalogDefs.py
+++ b/python/desc/twinkles/twinklesCatalogDefs.py
@@ -1,6 +1,7 @@
 """Instance Catalog"""
 from __future__ import absolute_import, division, print_function
 import numpy
+import numpy as np
 from lsst.sims.utils import SpecMap, defaultSpecMap
 from lsst.sims.catalogs.definitions import InstanceCatalog
 from lsst.sims.utils import arcsecFromRadians
@@ -39,9 +40,18 @@ class TwinklesPhoSimCatalogSN(PhoSimCatalogSN):
     def get_shorterFileNames(self):
         fnames = self.column_by_name('sedFilepath')
         sep = 'spectra_files/specFile_'
-        split_names = list(sep + fname.split(sep)[-1] for fname in fnames)
-        return split_names
+        split_names = []
+        for fname in fnames:
+            if 'None' not in fname:
+                fname = sep + fname.split(sep)[-1] 
+            else:
+                fname = 'None'
+            split_names.append(fname)
+        return np.array(split_names)
 
     column_outputs = PhoSimCatalogSN.column_outputs
     column_outputs[PhoSimCatalogSN.column_outputs.index('sedFilepath')] = \
         'shorterFileNames'
+
+    cannot_be_null = ['x0', 't0', 'z', 'shorterFileNames']
+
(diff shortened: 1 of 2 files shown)


TomGlanzman:
@rbiswas4, thank you for the updated script!  Two test visits are currently running and, so far, so good.  Will update as these tests complete.

An optimization question about the generatePhosimInput.py script.  There are a couple of web downloads, other than fatboy access:

Downloading http://maia.usno.navy.mil/ser7/finals2000A.all [Done]
Downloading http://sncosmo.github.io/data/models/salt2_extended.tar.gz [Done]

Do these files' content change frequently?  Is there any reason not to cache a copy locally to eliminate the ~25000 future downloads?

rbiswas4:
> Do these files' content change frequently? Is there any reason not to cache a copy locally to eliminate the ~25000 future downloads?

@TomGlanzman Those two files are for dustmaps and SNIa models. These files change with very low frequency, and in fact the design in `SNCosmo` was exactly what you suggest: At first use they are downloaded and then cached. My understanding was that this was still being followed in catsim, though the cache is one of the places where interaction with catsim is a problem.

I have a plan to investigate this separately but will probably not happen for the Twinkles run.

rbiswas4:
Please do not merge this yet: I tried to update the branch in order to resolve conflicts with master and I think I did something wrong. Unable to test anymore because I also think the server went down

rbiswas4:
I have the branch [`fix_merges`](https://github.com/DarkEnergyScienceCollaboration/Twinkles/tree/fix_merges)  which takes care of the merging issues (post the commit @TomGlanzman used) that are necessary to merge to master.

rbiswas4:
Yes, this is superseded by #373 
I have checked that I am getting the correct output on that.

jchiang87:
Ok, thanks.  I'm closing it to make it clear that it need not be considered any longer.

Issue #364, opened 2016-10-20
Create Strong Lensing Pipeline Review Notebook


jbkalmbach:
For the pipeline we want a strong lensing notebook that goes through several things:
- [ ] Explain Sprinkler
- [ ] Compare positions of lens images in Twinkles Instance Catalogs with positions from OM10 catalog
- [ ] Plot magnitude distributions, radial size and redshifts of lens galaxies in instance catalogs
- [ ] Plot magnitude distributions and redshifts of lensed AGN images in instance catalogs (Use brightest AGN image in a set)
- [ ] Check that variability is present in lensed AGN images in instance catalogs

jbkalmbach:
@drphilmarshall These are what I had in my notes from today's meeting, but if I'm missing things please add them.

drphilmarshall:
Looks good to me. Later we will need to explain how the sprinkler works: if it makes sense to do that now, please do go ahead and extend the checklist! Also, if you open the PR early I'll follow along and provide commentary as you push.

jbkalmbach:
Merged in #371

Issue #238, opened 2016-05-09
Overlay new coadds with sprinkled positions


jbkalmbach:
So, I compiled a list of sprinkled objects and then converted their ra,dec to pixel locations using the coadd wcs and astropy.wcs. 

First, I looked at the locations of all the sprinkled objects. Obviously, they cover a larger range than just the chip we are looking at.
![full_layout](https://cloud.githubusercontent.com/assets/4755090/15127435/8c77796c-15eb-11e6-89f8-ee1a2dbca541.png)

Then I looked just at the chip area.
![zoom_clean](https://cloud.githubusercontent.com/assets/4755090/15127436/8c7b3066-15eb-11e6-8984-4f1409291ae3.png)
![zoom_layout](https://cloud.githubusercontent.com/assets/4755090/15127438/8c7b30d4-15eb-11e6-8a80-a146f8aef764.png)

Then I picked up a few targets and zoomed in on them. It seems like they are all off just a little bit. My theories are that this is either due to something in the astropy wcs methods that may be different from the lsst dm methods or it's because the phosim catalogs are unrefracted coordinates and maybe this provides some sort of offset because it's not accounted for in the coadds? The green targets are the lens galaxies.
![zoom_targets_clean](https://cloud.githubusercontent.com/assets/4755090/15127439/8c7c9212-15eb-11e6-8eda-8f8c2922ef8d.png)
![zoom_targets](https://cloud.githubusercontent.com/assets/4755090/15127437/8c7b28e6-15eb-11e6-8244-c9abaf20b12e.png)

jbkalmbach:
Here I've attached the location information on the sprinkled objects and the lens galaxies.

[sprinkled_agn_pixel_locs.txt](https://github.com/DarkEnergyScienceCollaboration/Twinkles/files/255876/sprinkled_agn_pixel_locs.txt)
[sprinkled_lens_galaxy_pixel_locs.txt](https://github.com/DarkEnergyScienceCollaboration/Twinkles/files/255877/sprinkled_lens_galaxy_pixel_locs.txt)

SimonKrughoff:
Can I also get the ra, dec locations?

jbkalmbach:
Those files above should have ra,dec as well as the pixel coordinates I derived.

SimonKrughoff:
Following Bryce's lead, I plotted some positions of the SNe on the r-band coadd.
[...]

PR #371, opened 2016-10-24
Pipeline review notebook for Strong Lensing

jbkalmbach:
Need to add in a pipeline review notebook for the strong lensing aspects of the Twinkles simulation. The notebook should check off all the tasks outlined in #364


jbkalmbach:
@drphilmarshall I think this is a ready for some comments.

drphilmarshall:
Thanks @jbkalmbach ! The notebook format doesn't permit comments line by line, so here goes:

* The data files used look as though they are checked in, but with inaccurate filenames, I think. `phosim_galaxy_200.txt` should be in `data`, no? I guess the filename should inherit from its parent catalog in some way - and then if the `Sprinkler` produced these files, it should have added characters to the filename that show that the file contains sprinkled things.  **This needs to be done before merging this PR**

* Usually when the x-separations are off its because you need to multiply them by cos(dec). You are instead dividing the OM10 x separations by cos(dec) so it should work out, but it might be a good idea to do the comparison in arcsec with the OM10 numbers directly, converting the sprinkled positions down... This might remove some rounding error maybe? Not sure. Is cos(dec) about 0.9, by any chance? **This needs to be done before merging this PR**

* Some of the lens galaxies are going to pop out as being almost point sources - that'll be interesting :-)

* Distributions look sensible. It'd be nice to see a corner plot of z_lens, m_lens, z_agn, m_agn for both OM10 (gray) and Twinkles (red) to show that the Sprinkled objects match the OM10 distributions. Not urgent though.

* Agreed that we see variability - for the DESC Note we might quantify this with an RMS residual from the mean, perhaps? This could then be compared to the expected 0.01 mag photometric error.

* Again, for the DESC note we should explain how the lens population matching is done (ie with the XD model). But that's for later.

jbkalmbach:
On the first point, I was trying to avoid adding a full instance catalog ~60MB to the data file so I only took out the disk+bulge parts of the whole instance catalog for visit 200 and put them in phosim_galaxy_200.txt and put the AGN parts in phosim_agn_200.txt and uploaded them both as gists (that's where the links point). However, if we want them in the data folder, I could take only the lens galaxies that are sprinkled and only the AGN that we sprinkled in and add them as files in the data folder. Then they wouldn't be very large at all and I could tag them with 200 to signify they came from an instance catalog for visit 200.

jbkalmbach:
Yes, in the twinkles field we are around dec = -27.5 which gives cos(dec)~.89

drphilmarshall:
The only thing I can think of is that the sprinkler is computing RA incorrectly (without the cos(dec) factor). Here are the relevant lines from 

```python
# For each image, append the lens images
    for i in range(newlens['NIMG']):
        lensrow = row.copy()
        # XIMG and YIMG are in arcseconds
        # raPhSim and decPhoSim are in radians
        #Shift all parts of the lensed object, not just its agn part
        for lensPart in ['galaxyBulge', 'galaxyDisk', 'galaxyAgn']:
            lensrow[str(lensPart + '_raJ2000')] += np.radians(newlens['XIMG'][i]/(np.cos(np.radians(lensrow[str(lensPart + '_decJ2000')]))*3600.))
            lensrow[str(lensPart + '_decJ2000')] += np.radians(newlens['YIMG'][i]/3600.)
```

So, the cos(dec) is in there, correctly - but it looks like the lens RA and Dec are being treated inconsistently. In the last line, the declination is increased by a number in radians. However, in the line above, the declination is first converted to radians and then fed to the cosine function... I think we want:

```python
RA = lensrow[str(lensPart+'_raJ2000')]
Dec = lensrow[str(lensPart+'_decJ2000')]

dRA = np.radians( newlens['XIMG'][i] /3600.0 ) / np.cos(Dec)
dDec = np.radians( newlens['YIMG'][i] / 3600.0 )
            
lensrow[str(lensPart+'_raJ2000')] = RA + dRA
lensrow[str(lensPart+'_decJ2000')] = Dec + dDec
```

If the raJ2000 and decJ2000 quantities in the instance catalog are not already in radians, the first two lines would include a `np.radians()` conversion, and the last two would do an `np.degrees` (if that exists). What do you think?

jbkalmbach:
@drphilmarshall Talking with @rbiswas4 just now, he reminded me that the phosim instance catalogs have some intermediate transformations to get the right coordinates in the phosim images. See the discussion here: #238. Therefore, I need to make some new instance catalogs that are not phosim instance catalogs following @danielsf's [comment](https://github.com/DarkEnergyScienceCollaboration/Twinkles/issues/238#issuecomment-218232777)  in that thread.

jbkalmbach:
@drphilmarshall You were right about the problem in the sprinkler. The ra's are correct now. I apologize I didn't catch that sooner.

jbkalmbach:
@drphilmarshall This is ready for another look. There is the line in the code that fixes the ra positions that needs to be merged before @TomGlanzman can get the phosim runs going.

PR #373, opened 2016-10-26
Fix merges

### LINKED ISSUE #351: change to generatePhosimInputs.py to put only relative paths in instance catalogs

rbiswas4:
Testing the PR for the merge. (The previous branch was muddled in a merge)

The PhoSim Instance Catalogs outputs a string to the path of the spectra files relative to seddir argument supplied to the bin/generatePhoSimInstanceCatalog.py rather than the absolute path.

TwinklesSky now uses TwinklesPhoSimCatalogSN rather than PhoSimCatalogSN directly. 
TwinklesPhoSimCatalogSN inherits from PhoSimCatalogSN, but

defines a getter for shorterFileNames which is sedFilepath with prepended seddir stripped as requested.
redefine column_outputs to include shorterFileNames rather than sedFilepath
Fixes #351

### DIFF
--- a/python/desc/twinkles/twinkles_sky.py
+++ b/python/desc/twinkles/twinkles_sky.py
@@ -22,7 +22,7 @@ from lsst.sims.catUtils.exampleCatalogDefinitions import\
      DefaultPhoSimHeaderMap,
      DefaultPhoSimInstanceCatalogCols)
 from .twinklesCatalogDefs import TwinklesCatalogZPoint
-from .twinklesCatalogDefs import TwinklesPhoSimCatalogSN as PhoSimCatalogSN
+from .twinklesCatalogDefs import TwinklesPhoSimCatalogSN
 from desc.twinkles import (GalaxyCacheDiskObj, GalaxyCacheBulgeObj,
                            GalaxyCacheAgnObj, GalaxyCacheSprinklerObj,
                            create_galaxy_cache,
@@ -165,8 +165,8 @@ class TwinklesSky(object):
                              write_header=False)
 
         t_after_galCat = time.time()
-        snphosim = PhoSimCatalogSN(db_obj=self.snObj,
-                                        obs_metadata=self.obs_metadata)
+        snphosim = TwinklesPhoSimCatalogSN(db_obj=self.snObj,
+                                           obs_metadata=self.obs_metadata)
         ### Set properties
         snphosim.writeSedFile = True
         snphosim.suppressDimSN = True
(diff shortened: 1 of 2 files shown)


python/desc/twinkles/twinkles_sky.py
```diff
@@ -19,10 +19,10 @@
 from lsst.sims.catUtils.exampleCatalogDefinitions import\
     (PhoSimCatalogPoint,
      PhoSimCatalogSersic2D,
-     PhoSimCatalogSN,
      DefaultPhoSimHeaderMap,
      DefaultPhoSimInstanceCatalogCols)
 from .twinklesCatalogDefs import TwinklesCatalogZPoint
+from .twinklesCatalogDefs import TwinklesPhoSimCatalogSN as PhoSimCatalogSN
```
danielsf: I would rather import this as TwinklesPhoSimCatalogSN and then change the use of PhoSimCatalogSN to TwinklesPhoSimCatalogSN, so that, when looking at the body of the code, it is obvious that we are not using the sims_catUtils version of PhoSimCatalogSN

rbiswas4:
FYI @danielsf the change you requested has been made.

danielsf:
Okay.  Looks good.  You should probably go ahead and re-declare `column_outputs` to fix the order of the extinction parameters (internal first; Milky Way second).

I have a PR to do that here

https://github.com/DarkEnergyScienceCollaboration/Twinkles/pull/380

which re-declares `TwinklesPhoSimCatalogSN`.  I can just remove my class in favor of yours.

PR #374, opened 2016-10-26
Initialize scripts for pipeline task TW-phoSim-r3.  

TomGlanzman:
Used in first 40 test runs (almost true).  Note that the only change in this request is the addition of a new directory tree in Twinkles/workflows. It is expected this directory will be updated at least one more time prior to production of Run 3.


workflows/TW-phoSim-r3/phoSimPrep.py
```diff
@@ -0,0 +1,79 @@
...
+##       SCRIPT to generate phoSim instance catalog and custom SED library
+##
+# generatePhosimInput.py <opsHistID> --OpSimDBDir <loc-opsSim database> --seddir <loc-to-write-output SEDs> --outfile <instanceCat-name>
+### e.g., (the SEDs will occupy the "spectra_files" directory)
+# generatePhosimInput.py 200 --OpSimDBDir /nfs/farm/g/desc/u1/data/Twinkles --seddir . --outfile phosim_input_200.txt
+
+genCmd = 'generatePhosimInput.py'
+
+print '\nGenerate InstanceCatalog and SED files'
+cmd = "time "+genCmd+" "+obsHistID+" --OpSimDBDir /nfs/farm/g/desc/u1/data/Twinkles --seddir "+destSEDdir+" --outfile "+destIC
```
jchiang87: With the merges into master of #380 and #381, the calling sequence for `generatePhosimInput.py` has changed.  To take advantage of the local caches of the star, galaxy, and SN dbs, the `--cache_dir` option should be added:
```
 generatePhosimInput.py 230 --OpSimDBDir /nfs/farm/g/desc/u1/data/Twinkles --seddir <seddir> --outfile <outfile> --cache_dir /nfs/farm/g/desc/u1/data/Twinkles/fatboy_caches 
```
In this example, I've put the current locations of the `minion_1016_sqlite.db` file (`/nfs/farm/g/desc/u1/data/Twinkles`) and the `*_cache.db` files (`/nfs/farm/g/desc/u1/data/Twinkles/fatboy_caches`).  If those files are copied to local scratch space to avoid collisions with other processes trying to access them at the same time, those paths should of course be updated.

workflows/TW-phoSim-r3/helpers/updateGit.sh
```diff
@@ -0,0 +1,15 @@
...
+pipelineRoot=/nfs/farm/g/desc/u1/Pipeline-tasks/${taskName}
+
+GLOBIGNORE="*"
+src="${pipelineRoot}/config/"
+dst=${pipelineRoot}/Twinkles/workflows/${taskName}/
+
+echo src "${src}"
+echo dst ${dst}
+
+cmd="rsync -avuzb  --exclude ""*~"" ${src} ${dst}"
```
jchiang87: I think the copying here is going the opposite direction of what I would expect.  The files in the git repository should be your reference versions and they should be copied __to__ the execution area, not the other way around.  In fact, if I were doing this, at least for development, I would make symlinks in execution area that point back to the files in the repository, that way any edits I make in the repo are automatically available when I try to run the code.

Also, it's not clear to me that everything that ends up in the execution area should be copied back to the repository.  I'm thinking, in particular, of `twinkles_visits.txt` which would be created by running `get_twinkles_visits.py`.  Since that's generated by our code, it's really a data product of the execution of the data generation pipeline and should be put into the Data Catalog with metadata identifying it as being used in Run3, for example.

TomGlanzman: Re: reference copy      Thanks, excellent suggestion.  Done.

Re: small data files      Agreed.  But in the interests of getting the show on the road, I'd like to postpone this until the next go 'round.

TomGlanzman:
This was all done in something of a last minute rush, but I think it is ready for merging.  But I do not think I can do this as it means approving - which I am not allowed to do.  If this can be done by Monday, I will then begin a limited production of a few dozen visits while a final validation happens.

drphilmarshall:
Thanks Tom! Massive effort, much appreciated.

PR #380, opened 2016-10-31
write InstanceCatalogClasses to fix the order of rest/lab frame dust parameters

danielsf:
This should fix the order of extinction parameters in the PhoSim InstanceCatalogs (of course, we won't be able to test it until fatboy comes back on-line....)

### DIFF
--- a/python/desc/twinkles/twinkles_sky.py
+++ b/python/desc/twinkles/twinkles_sky.py
@@ -90,11 +90,11 @@ class TwinklesSky(object):
         # Lists of component phosim Instance Catalogs and CatalogDBObjects
         # Stars
         self.compoundStarDBList = [StarCacheDBObj]
-        self.compoundStarICList = [TwinklesPhoSimCatalogPoint]
+        self.compoundStarICList = [TwinklesCatalogPoint]
 
         # Galaxies
         self.compoundGalDBList = [GalaxyCacheBulgeObj, GalaxyCacheDiskObj, GalaxyCacheAgnObj]
-        self.compoundGalICList = [TwinklesPhoSimCatalogSersic2D, TwinklesPhoSimCatalogSersic2D,
+        self.compoundGalICList = [TwinklesCatalogSersic2D, TwinklesCatalogSersic2D,
                                   TwinklesCatalogZPoint]
 
         if not os.path.exists(_galaxy_cache_db_name):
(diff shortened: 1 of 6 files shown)


python/desc/twinkles/twinkles_sky.py
```diff
@@ -106,11 +103,11 @@ def __init__(self,
         # Lists of component phosim Instance Catalogs and CatalogDBObjects
         # Stars
         self.compoundStarDBList = [StarObj, CepheidStarObj]
-        self.compoundStarICList = [PhoSimCatalogPoint, PhoSimCatalogPoint]
+        self.compoundStarICList = [TwinklesPhoSimCatalogPoint, TwinklesPhoSimCatalogPoint]
```
jchiang87: Seems like this should be
```
        self.compoundStarICList = [TwinklesCatalogPoint, TwinklesCatalogPoint]
```

python/desc/twinkles/twinkles_sky.py
```diff
@@ -106,11 +103,11 @@ def __init__(self,
...
 
         # Galaxies
         self.compoundGalDBList = [GalaxyCacheBulgeObj, GalaxyCacheDiskObj, GalaxyCacheAgnObj]
-        self.compoundGalICList = [PhoSimCatalogSersic2D, PhoSimCatalogSersic2D,
+        self.compoundGalICList = [TwinklesPhoSimCatalogSersic2D, TwinklesPhoSimCatalogSersic2D,
```
jchiang87: should be
```
        self.compoundGalICList = [TwinklesCatalogSersic2D, TwinklesCatalogSersic2D,
```
?

python/desc/twinkles/twinkles_sky.py
```diff
@@ -165,8 +162,8 @@ def writePhoSimCatalog(self, fileName):
                              write_header=False)
 
         t_after_galCat = time.time()
-        snphosim = PhoSimCatalogSN(db_obj=self.snObj,
-                                        obs_metadata=self.obs_metadata)
+        snphosim = TwinklesPhoSimCatalogSN(db_obj=self.snObj,
```
jchiang87: Another one: `TwinklesCatalogSN` ?

python/desc/twinkles/twinkles_sky.py
danielsf: yup.  This is why you should never code when you can't also test...

jchiang87: I was able to run this code after merging into `u/danielsf/cache_stars`, but there were a number of conflicts I had to iron out.  It seems that this PR, #373, and #381 all touch the same code.   Is there a preferred order to merge all of these into master?

danielsf: I suggest @rbiswas4 should merge #373.  I will then pull down the new master, rebase, and fix any conflicts in this PR.  You can then re-review it before I merge.  Sound good?

danielsf:
Okay.  This branch is now in a state where it includes #373, as well as the cached db changes.  Do you want to have one more look before I merge, @jchiang87?

danielsf:
I also added a parameter `cache_dir` to the arg parser so that we can specify the directory where the database caches reside

jchiang87:
Yes, I'll have a look.

PR #384, opened 2016-11-04
Added sorting to sprinkler choices.

### LINKED ISSUE #379: Add in line to sort AGN matches in sprinkler by unique ID
@rbiswas4 brought up in discussion today that if we have multiple matches in the sprinkler that the database may not be returning results in the same order everytime so seeding the random choice is not sufficient to guarantee we always pick the same OM10 system for sprinkling. A quick one line solution to sort the AGN matches before the choice will fix this.

jbkalmbach:
To close #379 we want to sort before choosing which om10 system to accept in the sprinkler.

### DIFF
--- a/python/desc/twinkles/sprinkler.py
+++ b/python/desc/twinkles/sprinkler.py
@@ -121,6 +121,8 @@ class sprinkler():
                 if ((len(candidates) > 0) and (pick_value <= self.density_param)):
                     # Randomly choose one the lens systems
                     # (can decide with or without replacement)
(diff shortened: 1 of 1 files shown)


drphilmarshall:
I assume the sorting operation does not cause any knock-on ill effects, right?

jbkalmbach:
No, it doesn't change the OM10 data or the catsim data. It just makes sure we're going to have the same lenses in every instance catalog.

drphilmarshall:
Cool - thanks @jbkalmbach

PR #385, opened 2016-11-06
Fixed bug reported by @TomGlazman, and added extra print statement

rbiswas4:
- Bug was due to `Twinkles.3p2` being printed twice in place of
  `Twinkles.3p3`
- Added a print statement on how many viits are left out due to
  maxPredTime
	modified:   get_twinkles_visits.py

Fixed the bug discussed in #347 
- Added the `RF_pickle.p` file to `Twinkles/data` which was missing as reported by @TomGlanzman too in the same thread.


TomGlanzman:
@rbiswas4 I am now getting 21973 - 3 visits with 

Left out 2109 visits due tothe time limit of 100.0

for a grand total of 24,079.  A familiar number.

Thanks again!

PR #392, opened 2016-11-11
U/danielsf/validate centroid

danielsf:
This PR adds a script `compare_centroids.py` which reads in a centroid file and an InstanceCatalog.  It uses CatSim to predict the pixel positions of the sources in the InstanceCatalog.  It then generates a `.tex` file containing some scalar values and plots that summarize the agreement between the PhoSim pixel positions and the CatSim-predicted pixel positions of the sources in the InstanceCatalog.  This `.tex` file can be compiled into a `.pdf` using `pdflatex`.


danielsf:
@TomGlanzman I added a method to return a dict of scalar summary statistics.  Let me know if you want anything else.

Issue #396, opened 2016-11-13
Utility for examining diffims


SimonKrughoff:
Requirements:
specify visit on command line.
list possible diffims
Overlay masks
overlay dia sources -- red plus
overlay direct sources -- green circle
Display:
 Frame 0: Direct
 Frame 1: diff
 Frame 2: diff - direct

rbiswas4:
Somethings that we may want to do by adding to this:
- Can we later access the template and Direct (I am assuming this is the science image)  at different stages of processing? I am guessing that the default will be the template and science images transformed by (preconvolution, warping, and kernel matching) so that a direct pixel by pixel subtraction gives the resultant difference image?
- Can we change the threshold of detection?
- Can we look at the difference image around a ra, dec value?

SimonKrughoff:
> Can we later access the template and Direct (I am assuming this is the science image) at different stages of processing? I am guessing that the default will be the template and science images transformed by (preconvolution, warping, and kernel matching) so that a direct pixel by pixel subtraction gives the resultant difference image?

There are a couple of pre-defined data products that will help with this, but mostly this will require turning on the debugging output and running in the interactive debug mode.

> Can we change the threshold of detection?

Not interactively in a tool like this.  We would have to rerun the data to get this.

> Can we look at the difference image around a ra, dec value?

You mean other than using the built in utilities in DS9?  We could do this, but it seems like just using the display tools is the way to do this rather than building an ad hoc cutout service.

rbiswas4:
> Can we change the threshold of detection?
Not interactively in a tool like this. We would have to rerun the data to get this.

More clearly, I meant that after running the diffim, can I find out unique locations of a residual which have different levels of significance than the default. If that requires re-running, I  am guessing that you mean re-running the difference image task rather than the previous image based steps.

SimonKrughoff:
> More clearly, I meant that after running the diffim, can I find out unique locations of a residual which have different levels of significance than the default. If that requires re-running, I am guessing that you mean re-running the difference image task rather than the previous image based steps.

Correct.  There is no way to change the detection threshold interactively.  The diffim step needs to be rerun with another threshold.  There has been talk of allowing multiple detection thresholds, but I don't know how that translates into the mask layer.

rbiswas4:
Ok. But we can always run this looping through the diffim task at multiple thresholds if we really want to ...

SimonKrughoff:
Merged as part of #397

PR #397, opened 2016-11-13
Add a utility to look at diffim repositories

SimonKrughoff:
This is related to #396

### DIFF
--- a/bin/view_diffs.py
+++ b/bin/view_diffs.py
@@ -4,8 +4,11 @@ import lsst.daf.persistence as daf_persistence
 import lsst.afw.display as afw_display
 
 def list_diffims(butler):
-    """Use the butler to figure out what data are available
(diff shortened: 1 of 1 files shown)


SimonKrughoff:
> One of the things we have talked about is the need for providing test data for unit tests for such codes. No such data exists in Twinkles at the moment, so unit tests are not possible. Do you think we could make a data repository with three sources for a coadd after which we could do such a test for code such as this, or is the requirement for such code steeper?

The problem is that I don't know how to unit test code that does essentially only display.  If there are thoughts on this, I'd love to hear them.  I suppose we could unit test the failure modes, but that is a very minimal aspect of what this does.

SimonKrughoff:
I have changed up the docs.

I captured the need to move the free functions into the desc.twinkles namespace in #404.

I don't think I know how to unit test this at the moment, but I'm open to suggestions.

PR #418, opened 2016-11-23
Twinkles 3p4 : the visits that should be part of Twinkles but were left

### LINKED ISSUE #416: Defining a Run 3.4 list for long-CPU time visits
Run 3 has 2109 visits that are not part of the planned sets of phosim runs (designated 3.1, 3.1b, 3.2, or 3.3) because they are anticipated to require more than 100 CPU hours each.  Phil recommends that we ask @rbiswas4 to assemble a '3.4' list of these visits for potential future reference.

rbiswas4:
out of Twinkles 3p1-3 due to estimates of PhoSim run times being too
long

- Added code to define the visits for Twinkles 3p4 split into two tuple
  elements for WFD, and DDF
	modified:   python/desc/twinkles/obsHistIDOrdering.py
- Modified the script to generate these visits
	modified:   bin/get_twinkles_visits.py
- Modified example notebook to demonstrate this
	modified:   examples/notebooks/Example_OpSimOrdering.ipynb
Fixes #416

### DIFF
--- a/bin/get_twinkles_visits.py
+++ b/bin/get_twinkles_visits.py
@@ -53,6 +53,15 @@ if args.orderObsHistID:
     with open(args.outfile, 'a') as output:
         output.write('# Begin Section Twinkles 3.3\n')
     ops.Twinkles_3p3.obsHistID.to_csv(args.outfile, index=False, mode='a')
+
+    Obs_3p4a, Obs_3p4b = ops.Twinkles_3p4
+    with open(args.outfile, 'a') as output:
+        output.write('# Begin Section Twinkles 3.4 WFD\n')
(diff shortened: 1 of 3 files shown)


rbiswas4:
Since the Twinkles runs are crunching towards an end, here is Twinkles 3.4. The csv file can be generated by running the `bin/get_twinkles_visits.py` script in the same way as before. The Twinkles 3p4 visits are added to the csv file with additional headers for the WFD and DDF visits.

rbiswas4:
The dataframe is populated through a sql query which is ordered by `PROPID`.  So, while the outputs have appeared to be sorted in mjd whenever I have looked, I don't know of a good reason why this must be the case in a database. Eitherway, this guarantees the ordering we want.

PR #430, opened 2017-01-18
Remove patch and tract from templateId argument

SimonKrughoff:
This was a typo in the original cookbook.  I didn't have those in my original notes.  They cause a crash when they are present.


drphilmarshall:
PS. I see the coverage has gone down - a unit test to ensure we don't get the patch and tract wrong again could be good?

PR #442, opened 2017-03-02
Created an Instance Catalog parsing notebook.

jbkalmbach:
Based on work from the hack day I created a way to read an instance catalog and split the sprinkled objects out into their own pandas dataframes. I think this could be useful and should be added. Maybe it should go into documentation instead of examples?


jbkalmbach:
@linan7788626 @mpwiesner Here is a notebook that I started on hack day which will let you find the sprinkled objects (such as the host galaxies) in a Twinkles Instance Catalog. Hasn't been merged yet, but wanted to let you guys know about it.

mpwiesner:
Thanks, Bryce!



Matthew P. Wiesner, Ph.D.
Website http://nicadd.niu.edu/~mwiesner
Twitter: @matthewpwiesner
Blog http://quantumefficiency.wordpress.com




-----Original Message-----

linan7788626:
Hi @jbkalmbach , thank you so much for the update, I am thinking of combining modules together, and will visit @mpwiesner to push things forward. We will keep you updated. 

I have one question about the hosts of quasars: in principle, the number of hosts of lensed quasars should be the same as the number of lensing systems, i.e., even though where are 4 lensed quasars, but there is only one host of these quasars (the host of the source quasar). But when I check the catalog you sent us during the hackday, it seems that there are more hosts than lensing systems. Could you please tell me how you create this catalog briefly? Thank you so much!

jbkalmbach:
This is explained in the end of the notebook I've created, but it's because some host galaxies have both bulge and disk components, but not all do. Catsim creates catalogs for galaxies from separate bulge, disk and AGN components.

linan7788626:
Hi Bryce,

Great, I understand now, thanks a lot. 

Best,
Nan Li

linan7788626:
Hi @jbkalmbach 

I checked the instance catalogs of the host galaxies and tried to match them to the sprinkled AGNs. But after looking into the catalogs, I am still confused by the relation between the hosts of the source AGNs and the sprinkled AGNs. My confusion is similar to the question I asked before; even I took the number of components of the hosts into account, it does not make to me either. I will show you an example to explain my concerns (The tables below would be bigger than it is shown):

Let us focus on the system of  TwinkleID = 24 (hereafter T24). The instance catalogs of the sprinkled AGNs in T24 should be:

galid | ra | dec | mag | sed_name | z | catsim_id | twinkles_system | twinkles_img_num  
---|---|---|---|---|---|---|---|---
213934178332 | 53.052941 | -27.7033144 | 21.1014594 | agnSED/agn.spec.gz | 0.48 | 208920096 | 24 |0
213934179356 | 53.0523965 | -27.7029225 | 22.6117574 | agnSED/agn.spec.gz | 0.48 | 208920097 | 24 | 1

As you mentioned, to match the host galaxy of the source AGN of these sprinkled AGNs, we need to match the CatSimID of the AGNs here to the CatSimIDs in the file of "sprinkled_hosts_230.txt".
Then I found these parameters:

galid | ra | dec | mag | sed_name | z | spatialmodel | majorAxis | minorAxis | positionAngle | sindex | catsim_id
---|---|---|---|---|---|---|---|---|---|---|---
213934178330 | 53.052941 | -27.7033144 | 19.2833768 | galaxySED/Inst.50E09.04Z.spec.gz | 0.48 | sersic2d | 1.00133634 | 0.981526315 | 306.387299 | 4 | 208920096   
213934178331 | 53.052941 | -27.7033144 | 25.0634271 | galaxySED/Burst.16E09.04Z.spec.gz | 0.48 | sersic2d | 3.89396667 | 3.36492896 | 306.387299 | 1 | 208920096   
213934179354 | 53.0523965 | -27.7029225 | 20.77273 | galaxySED/Inst.50E09.04Z.spec.gz | 0.48 | sersic2d | 1.00133634 | 0.981526315 | 306.387299 | 4 | 208920097   
213934179355 | 53.0523965 | -27.7029225 | 26.5527803 | galaxySED/Burst.16E09.04Z.spec.gz | 0.48 | sersic2d | 3.89396667 | 3.36492896 | 306.387299 | 1 | 208920097 

In this table, "line 1 + line 2" represents the host of image 0 in T24, and "line 3 + line 4" represents the host of image 1 in T24. Theoretically, the sprinkled AGNs are lensed from the same source AGN, so the host galaxies of image 0 and image 1 in T24 should be exactly the same, however, if you check the magnitudes of the corresponding components of each lensed AGN, you can see they are different. 

I suppose that you generated these hosts according to the lensed AGNs but not the source AGN,  and this approach might be incorrect ( in my option). The proper way to generate the host of the source AGN should be: 

- Find the properties of the source AGN of a given Twinkle system (say T24) in the corresponding lensing system in OM10. 
- Create all the components of the host galaxy according to properties of the source AGN.

Do my concerns make sense to you, if I misunderstood anything or you have any questions, please feel free to let me know, I really appreciate it.
Thank you so much!

All the best,
Nan Li

jbkalmbach:
Hi Nan,

The host galaxies of the source AGN are also magnified by the same amount as the lensed AGN in the catalog. That's why there is a difference between the entries. Perhaps this isn't the best thing to do. What do you think about this? Also, @drphilmarshall thoughts on this? This isn't something we dealt with in Twinkles 1 and maybe should be moved to an actual issue.

linan7788626:
Hi @jbkalmbach, 

Thank you for your reply. You are right; this might be not the best thing to generate host galaxies because the lensing effects on AGNs and their host are different. It is probably not reasonable that simply magnifying the host by using the same magnifications of the lensed AGN, and this is the reason why we adopt ray-tracing to calculate the correct magnification of the lensed host galaxies. My idea about this is that: rescaling the host to the original one by subtraction the magnification of the lensed AGN, i.e. we need to recalculate the magnitude of the host according to the magnification of the corresponding AGN and the magnified magnitude of the host so far we have. I will think about the details of how to implement the rescaling of the hosts, and keep you updated.

Thanks,
Nan

Issue #449, opened 2017-05-17
Start an LSST DESC Note: The Twinkles 1 Strong Lens Population


drphilmarshall:
@jbkalmbach I'll start an `ipynb` Note on this for us. A lot of the material in the [MatchingLensGalaxies.ipynb](https://github.com/LSSTDESC/Twinkles/blob/master/examples/notebooks/MatchingLensGalaxies.ipynb) example is relevant, but we also need to introduce, explain and demonstrate the sprinkler code - preferably all the way through to making the actual instance catalog used in Run 3. I can suggest some section headings before passing this on to you - sound good?

drphilmarshall:
OK, Note is being developed in PR #450!

PR #450, opened 2017-05-17
LSST DESC Note: The Twinkles 1 Strong Lens Population

drphilmarshall:
@jbkalmbach Here's an ipython notebook for us to write up the SL inputs work (showing the sprinkler in action, and the lens matching etc, #449). Can you take a look and maybe refine the section headings, do some subsectioning etc? 

It looks to me as though we might need to refactor the sprinkler code so that it handles the lens galaxy matching - that stuff is just in an `example` notebook right now. What do you think?

We won't merge this PR until we've finished writing the Note and the group has reviewed it.


jbkalmbach:
The lens galaxy matching shouldn't be a part of the sprinkler since we just run the example notebook once and create a new fits catalog of lens systems with the additional information we have added about the lens galaxies. At that point the sprinkler just uses this new catalog whenever you run it. I think we should certainly mention it in the mock catalog section of the notebook and maybe even summarize the example notebook adding some of the plots from there, but I don't think the code belongs in the sprinkler.

drphilmarshall:
OK: I guess the Note should explain this, and instead present the matching
as providing necessary additional information to what is present in the
OM10 catalog. Thanks!

jbkalmbach:
This note is ready for comments. The Discussion and Conclusions section is kind of a set of bullet point response right now and any comments on the content there would be helpful before I write it out.

PR #451, opened 2017-05-17
LSST DESC Note: The Twinkles1 PhoSim Pipeline and Results

drphilmarshall:
Hi @TomGlanzman! I started an LSST DESC Note for you, check it out: 

https://github.com/LSSTDESC/Twinkles/blob/twinkles1-phosim-pipeline-note/doc/LSST_DESC_Notes/desc-0000-twinkles-1-phosim-pipeline/main.rst

As you can see, the development branch for this note is called [`twinkles1-phosim-pipeline-note`](https://github.com/LSSTDESC/Twinkles/blob/twinkles1-phosim-pipeline-note/doc/LSST_DESC_Notes/). We won't merge this PR until you're happy with the Note, and the group has reviewed it. I don't think this needs to be a long Note, but it should include listings of the commands that you used to execute the Run 3 image simulation jobs, I think - and then some results  of the kind you've been posting to the issue threads (on job success rate, wallclock vs CPU time, long jobs, and so on). What do you think?


TomGlanzman:
@drphilmarshall That all sounds very reasonable.

drphilmarshall:
Rock on! Feel free to use this thread to ask me, @danielsf, @tony-johnson or anyone else for input / help.

PR #462, opened 2018-01-25
Issue/461/proto dc2 nersc

jbkalmbach:
Updates for DC2 Run 1.1

Major additions:
* Added Sprinkled Sne functionality
* Added ability to read cached gal_ids
* Corrected magnorm for AGN SEDs


jbkalmbach:
I added @rbiswas4 as a reviewer for the SN additions and @danielsf for the other code which is mostly subclassing existing sims packages. If you guys don't have time we can find somebody else to help as well.

jbkalmbach:
Also, we should wait to look at this until the sprinkler is working with sims_GCRCatSimInterface.

python/desc/twinkles/twinklesCatalogDefs.py
```diff
@@ -63,7 +78,109 @@ def get_shorterFileNames(self):
...
                       'phoSimMagNorm', 'shorterFileNames', 'redshift',
-                      'shear1', 'shear2', 'kappa', 'raOffset', 'decOffset',
+                      'gamma1', 'gamma2', 'kappa', 'raOffset', 'decOffset',
                       'spatialmodel', 'internalExtinctionModel',
                       'galacticExtinctionModel', 'galacticAv', 'galacticRv']
     cannot_be_null = ['x0', 't0', 'z', 'shorterFileNames']
+
+class TwinklesCompoundInstanceCatalog(CompoundInstanceCatalog):
+
+    def write_catalog(self, filename, chunk_size=None, write_header=True, write_mode='w'):
```
danielsf: How is this different from `CompoundInstanceCatalog.write_catalog()`?

python/desc/twinkles/twinklesDBConnections.py
```diff
@@ -38,3 +39,52 @@ class SNCacheDBObj(SNDBObj, CatalogDBObject):
 
     def query_columns(self, *args, **kwargs):
         return CatalogDBObject.query_columns(self, *args, **kwargs)
+
+class TwinklesBulgeObj(GalaxyBulgeObj):
+
+    database = 'LSSTCATSIM'
```
danielsf: Should we just define these connection parameters in a mixin that all of the `Twinkles*Obj` classes can inherit from?

python/desc/twinkles/twinklesCatalogDefs.py
jbkalmbach: In line 179-180 it adds:
```
 +                compound_dbo.mjd = self._obs_metadata.mjd.TAI
 +                compound_dbo.specFileMap = twinkles_spec_map
```
which the sprinkler needs access to.

python/desc/twinkles/twinklesDBConnections.py
```diff
@@ -38,3 +39,52 @@ class SNCacheDBObj(SNDBObj, CatalogDBObject):
...
+class TwinklesAgnObj(GalaxyAgnObj):
+
+    database = 'LSSTCATSIM'
+    port = 1433
+    host = 'fatboy.phys.washington.edu'
+    driver = 'mssql+pymssql'
+
+    objid = 'galaxyAgn'
+
+    columns = [('galtileid', None, numpy.int64),
```
danielsf: This `columns` is just the one that is defined in `GalaxyAgnObj`.  There shouldn't be a need to re-define it here.

Is all of this reproduced code a hold-over from Twinkles 1?  Were we running against an out-dated version of `lsst_sims` and needed to re-implement `master` in the Twinkles repo?

python/desc/twinkles/twinklesCatalogDefs.py
danielsf: Right. Now I remember talking about that.

Now that I see this solution in practice, it makes me uneasy. Could we just add a line to DESCQAObject.query_columns that sets self.obs_metadata = obs_metadata. This would give you access to MJD once the query is run. As far as the twinkles_spec_map, could that be handled at the InstanceCatalog level, the way that we handle the full file path mapping in "normal" PhoSim InstanceCatalogs

https://github.com/lsst/sims_catUtils/blob/master/python/lsst/sims/catUtils/exampleCatalogDefinitions/phoSimCatalogExamples.py#L220

?

danielsf: Or should we add a `query_metadata` dict to `DESCQAObject` that can contain `obs_metadata`?  That way it is clear(er) that this is data associated with a specific query, rather than data associated with the `DESCQAObject` in any permanent sense.

examples/notebooks/set_x0_sn_catalog.ipynb
```diff
@@ -0,0 +1,1338 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 49,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "cosmo = FlatLambdaCDM(H0=71, Om0=0.265, Tcmb0=0, Neff=3.04, m_nu=None, Ob0=0.0448)"
```
rbiswas4: Thanks Bryce ! Can you please add this as a variable like `DC2Cosmo` so that this is not hard coded. 
Also, it is likely that the cut paste I did on slack did not work accurately.

jbkalmbach:
@rbiswas4 Thanks for checking it out! To answer your question, I do set the ra, dec [here](https://github.com/LSSTDESC/Twinkles/blob/issue/461/proto_dc2_nersc/python/desc/twinkles/sprinkler.py#L381-L382)

rbiswas4:
@jbkalmbach It is reasonable to expect that what you did should work, but unfortunately it does not :( . You need an additional line 
```
current_sn_obj = self.sn_obj.fromSNState(sn_param_dict) # in your code
current_sn_obj.mwEBVfromMaps()
```
So, you should add that in your code for now. 

Later, on this is something that we should discuss and come up with ways of improving, because of the ability of causing confusion. The reason for requiring this extra line of code was to allow people the facility of setting the `EBV` value independently of a dustmap and coordinates, which is sometimes useful, and there is a discussion of this in the `SNObject` example.  Maybe we can redesign it to have this feature kick on after setting a suitable attribute?

jbkalmbach:
Ah, ok. I will add that line.

jbkalmbach:
Merging and will add an issue to go over the code and refactor after Run 1.1 is underway.

PR #464, opened 2018-01-31
Updated dc2 caches for correct uDDF area.

jbkalmbach:
New AGN Locations inside uDDF from Run 1.1 Google Doc
agn_cache
New SNe Locations
sne_cache


jbkalmbach:
Ok updated plots:
agn_cache
sne_cache

jbkalmbach:
And just so we have it somewhere this gives 1028 sprinkled AGN and 953 sprinkled SN

PR #469, opened 2018-03-29
Run1.2 update

jbkalmbach:
* Updated catalog caches for DC2 Run 1.2.
* Silenced output from OM10 and Sprinkler


jchiang87:
@jbkalmbach The code changes look ok to me, but when I try to run using this branch, the `u/danielsf/debug/dust_columns` branch of `sims_GCRCatSimInterface`, and the `v0.7.1` tag of `gcr-catalogs` on visit 197356, which contains the uDDF region, I'm not getting any point sources in the `gal_cat_197356.txt` catalog for a 2.1 deg radius fov.   Can you try running that visit with all of the updates for Run1.2 and see if you can confirm?

jbkalmbach:
Hi @jchiang87, It looks like that tag was set before the oversize `galaxy_ids` were changed in protoDC2_v3 so maybe we can ask @yymao to issue a new tag on gcr-catalogs?

yymao:
@jchiang87 @jbkalmbach done: https://github.com/LSSTDESC/gcr-catalogs/releases/tag/v0.7.2

jbkalmbach:
Thanks @yymao!

jchiang87:
Thanks, both. Using that tag seems to have addressed the issue. I'll report back shortly once the run has finished.

jchiang87:
Here's a plot of galaxies (2D histogram) and sprinkled (and some non-sprinkled) AGN for visit 197356 for the protoDC2 field using the protoDC2_v3 data.
sprinkled_v197356_2 1deg-fov_full_wfd
My `generateInstCat.py` run is still going, but I think this is good to go.

PR #471, opened 2018-04-02
Increase GLSN redshift range for redshifting of SEDs.

jbkalmbach:
This should be a quick fix for Run 1.2 redshifting issues.


jbkalmbach:
Great! I'll merge this at 2pm PDT if nobody has any other comments.

PR #472, opened 2018-05-21
U/danielsf/sprinkled/truth

danielsf:
This PR is in support of a similar PR being made to `sims_GCRCatSimInterface`.  It wires the sprinkler to add an `is_sprinkled` column, which is set to 1 for objects changed or added by the sprinkler and 0 for every other object.

I believe I have set this up so that the sprinkler will only try to set `is_sprinkled` if it is already present, so we shouldn't have to change any other, non-DC2 code that depends on the sprinkler.


danielsf:
@jbkalmbach I think I'm going to need you to officially approve this before I can merge it.

PR #473, opened 2018-07-02
Issue/467/ic validation

jbkalmbach:
Instance catalog validation tests. Not quite ready for full review but ready for discussion.


jbkalmbach:
Ok, this is ready for a look. At the moment it raises errors when a test fails but if there is another action we should take we can discuss and make changes.

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+
+    def process_hosts(self, sprinkled_df, df_galaxy):
+
+        host_bulge_locs = []
+        host_disk_locs = []
+        for bulge_idx, disk_idx in zip(sprinkled_df['host_galaxy_bulge_uID'],
+                                       sprinkled_df['host_galaxy_disk_uID']):
+            bulge_matches = np.where(df_galaxy['uniqueId'] == bulge_idx)[0]
+            disk_matches = np.where(df_galaxy['uniqueId'] == disk_idx)[0]
+            for bulge_match, disk_match in zip(bulge_matches, disk_matches):
```
danielsf: Do you need to verify that `len(bulge_matches) == len(disk_matches)`?  `zip` will let you combine arrays of different lengths, truncating the longer of the two.

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+                host_bulge_locs.append(bulge_match)
+                host_disk_locs.append(disk_match)
+
+        host_bulges = df_galaxy.iloc[np.unique(host_bulge_locs)]
+        host_bulges = host_bulges.reset_index(drop=True)
+        host_bulges['galaxy_id'] = sprinkled_df['galaxy_id']
```
danielsf: Is there any risk that `len(host_bulges) != len(sprinkled_df)`?

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+
+        lens_gals = df_galaxy.iloc[np.unique(lens_gal_locs)]
+        lens_gals = lens_gals.reset_index(drop=True)
+
+        return lens_gals
+
+    def process_sne_lenses(self, df_galaxy):
+
+        """
+        These must be done differently since the lens galaxy for SNe
```
danielsf: Can you explain what you mean by "differently"?

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+        keep_idx = []
+        for sne_idx, sne_row in df_sne.iterrows():
+            if sne_row['sedFilepath'].startswith('Dynamic'):
+                sne_id = sne_row['uniqueId']
+                twinkles_ids = np.right_shift(sne_id-117, 10)
+                galtileid = int(twinkles_ids/10000)
+                if galtileid in galtile_list['galtileid']:
+                    keep_idx.append(i)
+                    galtileids.append(galtileid)
+                    twinkles_num = np.int(str(twinkles_ids)[-4:])
```
danielsf: Same comment as above about using `twinkles_ids % 10000` instead of `np.int(str(...)[-4:])`

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+
+        galtileids = np.array(galtileids, dtype=np.int)
+        sprinkled_sne = df_sne.iloc[np.array(keep_idx)]
+        sprinkled_sne = sprinkled_sne.reset_index(drop=True)
+        sprinkled_sne['galaxy_id'] = galtileids
+        sprinkled_sne['twinkles_system'] = twinkles_system
+        sprinkled_sne['image_number'] = twinkles_im_num
+        sprinkled_sne['lens_galaxy_uID'] = np.left_shift(np.array(galtileids, dtype=np.int), 10) + 97
+        sprinkled_sne['host_galaxy_bulge_uID'] = np.array(np.left_shift(galtileids*10000
+                                                    + np.array(twinkles_system, dtype=np.int)*4 +
```
danielsf: Same comment as above about not casting `twinkles_system` and `twinkles_im_num` as numpy arrays more than once.

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+
+            spr_sys_df = spr_sne_df.query('lens_galaxy_uID == %i' % u_id)
+
+            # There may be no images present in the current instance catalog
+            if len(spr_sys_df) == 0:
+                continue
+
+            # This is just to make sure there are at least some images when we expect
+            # and the previous line is not just skipping everything.
+            images_tested += 1
```
danielsf: Should this be moved into the `for img_idx` loop below, just in case `spr_sys_df['image_number'].values == 0`?

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+                    errors_present = True
+                    lens_major_axis_error = True
+
+            if (lens_gal_df['minorAxis'] - lens['REFF']/np.sqrt(1+lens['ELLIP'])) > 0.005:
+                if lens_minor_axis_error is False:
+                    errors_string = errors_string + "\nLens Minor Axis. First error found in lens_gal_id: %i " % u_id
+                    errors_present = True
+                    lens_minor_axis_error = True
+
+            if (lens_gal_df['sedFilepath'] != lens['lens_sed'][0]):
```
danielsf: Do you mind adding a test on `os.path.isfile(lens_gal_df['sedFilepath'])` to make sure that the SED file exists?  I know that I added a commit a few weeks ago that breaks the SED filenames output by the sprinkler.  I will fix it once we have a viable copy of cosmoDC2 to test on.  I would sleep easier if I knew SED file existence was being tested on (unless, I guess, it proves to be too expensive).

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+        for lens_gal_row in spr_agn_lens_df.iterrows():
+
+            lens_idx, lens_gal_df = lens_gal_row
+            u_id = lens_gal_df['uniqueId']
+
+            spr_sys_df = spr_agn_df.query('lens_galaxy_uID == %i' % u_id)
+            use_lens = db.lenses['LENSID'][np.where(db.lenses['twinklesId'] == 
+                                                    spr_sys_df['twinkles_system'].iloc[0])[0]]
+            lens = db.get_lens(use_lens)
+
```
danielsf: Should all of these thresholds be in `np.abs()`? 

i.e. what if `lens['PHIE']+lens_gal_df['positionAngle'] = -10.0`?

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+            u_id = lens_gal_df['uniqueId']
+
+            spr_sys_df = spr_agn_df.query('lens_galaxy_uID == %i' % u_id)
+            use_lens = db.lenses['LENSID'][np.where(db.lenses['twinklesId'] == 
+                                                    spr_sys_df['twinkles_system'].iloc[0])[0]]
+            lens = db.get_lens(use_lens)
+            num_img = lens['NIMG']
+
+            test_sed = Sed()
+            test_sed.readSED_flambda('%s/%s' % (galDir, lens['lens_sed'][0]))
```
danielsf: Where is `lens['lens_sed']` coming from?  If it is coming from the InstanceCatalog, you should not need `galSpecDir` here (the InstanceCatalog should already refer to `galSpecDir`).  This may be the bug that I introduced with

https://github.com/LSSTDESC/sims_GCRCatSimInterface/commit/74a4f32b2d3e0b7c38b5a558dfd3d749b17adf26

(that commit takes the mapping with `sims.utils.defaultSpecMap` out of the getter method; I forgot that the sprinkler would not know about this change; I need to do some profiling to see if this really gets us a significant improvement in performance)

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+            # These should just be different by a minus sign
+            # Also making sure that this works for both entries in lens (should be the same anyway)
+            if np.max(lens['theta_e'] + lens_gal_df['positionAngle']) > 0.005:
+                if phi_error is False:
+                    errors_string = errors_string + "\nPosition Angles. First error found in lens_gal_id: %i" % u_id
+                    errors_present = True
+                    phi_error = True
+
+            # Redshifts should be identical
+            if np.max(lens['zl'] - lens_gal_df['redshift']) > 0.005:
```
danielsf: Again, should these error thresholds be in `np.abs()`?

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+            # There may be no images present in the current instance catalog
+            if len(spr_sys_df) == 0:
+                continue
+
+            lens = df.query('twinkles_sysno == %i' % spr_sys_df['twinkles_system'].iloc[0])
+            img_vals = spr_sys_df['image_number'].values
+
+            for idx, image_on in list(enumerate(img_vals)):
+                
+                if lens['zs'].iloc[image_on] - spr_sys_df['redshift'].iloc[idx] > 0.005:
```
danielsf: Should this error threshold be in `np.abs()`?

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+                    if z_s_error is False:
+                        errors_string = errors_string + "\nSNe Image Redshifts. First error found in gal_id: %i " % u_id
+                        errors_present = True
+                        z_lens_error = True
+
+                sed_name = '%s/specFileGLSN_%i_%i_%.4f.txt.gz' % (sne_file_loc,
```
danielsf: We should also check that the SNe SED files exist, since there is a risk that they won't get written out properly.

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+
+                d_mags = self.get_agn_variability_mags(agn_var_params[str(agn_id.values[0])],
+                                                       lens['DELAY'].data[0][i], visit_mjd,
+                                                       spr_sys_df['redshift'].values[0])
+
+
+                test_sed = Sed()
+                test_sed.readSED_flambda('%s/%s' % (agnDir, 'agn.spec.gz'))
+                test_sed.redshiftSED(lens['ZSRC'])
+                test_f_norm = test_sed.calcFluxNorm(lensed_mags.values[i], norm_bp)
```
danielsf: The method `lsst.sims.photUtils.getImsimFluxNorm` is generally faster than calling `calcFluxNorm` with the imsim bandpass.  I recommend using that here.

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+        var_mag_dict = {x:y for x,y in zip(filters, var_mags)}
+
+        return var_mag_dict
+
+        
+    def load_agn_var_params(self):
+
+        agn_var_params = {}
+
+        with open(os.path.join(os.environ['TWINKLES_DIR'], 'data',
```
danielsf: Can you re-generate `agn_validation_params.txt`, delineating the columns with `;`?  Then, you can just split on `;` and use `json.loads()` to get the parameter dicts, rather than having to munge the strings below.

python/desc/twinkles/validate_ic.py
```diff
@@ -0,0 +1,849 @@
...
+            spr_sys_df = spr_sne_df.query('lens_galaxy_uID == %i' % u_id)
+
+            # There may be no images present in the current instance catalog
+            if len(spr_sys_df) == 0:
+                continue
+
+            lens = df.query('twinkles_sysno == %i' % spr_sys_df['twinkles_system'].iloc[0])
+            img_vals = spr_sys_df['image_number'].values
+
+            mag = lens['mu']
```
danielsf: Given that you take the `log` of this below, I gather that this is a flux.  It is probably best to call it `flux`.

utils/run_validation.py
```diff
@@ -0,0 +1,67 @@
+import os
+import sys
+import matplotlib as mpl
+mpl.rcParams['text.usetex'] = False
+import matplotlib.pyplot as plt
+from desc.twinkles import validate_ic
+
+if __name__ == "__main__":
+
+    cat_folder = sys.argv[1]
```
danielsf: How is this script meant to be run?  If it is supposed to be run from the command line, can you use `argparse` to get the arguments so that `python utils/run_validation.py -h` gives some helpful information?

python/desc/twinkles/validate_ic.py
jbkalmbach: `lens['lens_sed']` comes from the sprinkled lens catalog (OM10 lens plus added info like SEDs from CatSim) so this is needed here. Plus, I need to load the actual SED so I need the full filepath to the stored copy of the `sims_sed_library`.

python/desc/twinkles/validate_ic.py
jbkalmbach: These SEDs should just be in the `sims_SED_library`. Do you think that's necessary to test?

python/desc/twinkles/validate_ic.py
jbkalmbach: Actually a magnification factor on the flux. Therefore I will label it `magnification` to be clearer.

python/desc/twinkles/validate_ic.py
danielsf: Yes.  The damage I did in `Twinkles` revolved around where the SED subdir (`galaxySED`) gets prepended to the filename.  In DC1, the InstanceCatalog class added `galaxySED` through a call to `sims.utils.defaultSpecMap`.  In order to speed up SED matching in DC2, I moved that step into the SED fitter.

https://github.com/LSSTDESC/sims_GCRCatSimInterface/commit/74a4f32b2d3e0b7c38b5a558dfd3d749b17adf26

python/desc/twinkles/validate_ic.py
danielsf: OK

jbkalmbach:
Ok, think we're good to go for another look.

jbkalmbach:
If nobody has any more comments I will merge this some time after 3 PM Pacific today.

PR #475, opened 2018-07-12
U/danielsf/truth/catalogs

danielsf:
This PR adds functionality I needed to generate the DC2 truth catalogs.  Mainly, that means adding code to allow access to SNe parameters, even when the SNe is not currently going off.

I also did a little bit of munging on the `galaxy_id` question to get it to work for protoDC2.  We will need a more permanent fix for CosmoDC2, as Bryce and I discussed a few days ago.

### DIFF
--- a/python/desc/twinkles/sprinkler.py
+++ b/python/desc/twinkles/sprinkler.py
@@ -73,7 +73,7 @@ class sprinkler():
             The results array from an instance catalog.
         visit_mjd: float
             The mjd of the visit
-        specFileMap: 
+        specFileMap:
             This will tell the instance catalog where to write the files
         om10_cat: optional, defaults to 'twinkles_lenses_v2.fits
             fits file with OM10 catalog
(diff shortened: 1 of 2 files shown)


python/desc/twinkles/sprinkler.py
```diff
@@ -242,7 +256,7 @@ def sprinkle(self):
                         #just use np.right_shift(phosimID-28, 10). Take the floor of the last
                         #3 numbers to get twinklesID in the twinkles lens catalog and the remainder is
                         #the image number minus 1.
-                        lensrow[self.defs_dict['galtileid']] = (lensrow[self.defs_dict['galtileid']]*10000 +
+                        lensrow[self.defs_dict['galtileid']] = ((lensrow[self.defs_dict['galtileid']]+30000000)*10000 +
```
jbkalmbach: Was this just to avoid clashes for now?

danielsf:
@jbkalmbach Can you check what I did in `f259cde`, to make sure you are okay with that?  In order to get all three versions of `galaxy_id` mapped correctly (bulge, disk, agn), I needed to change how `self.defs_dict` is loaded.

jbkalmbach:
Ah, I see. Yeah, looks fine. It will only be a tuple when using the truth methods, right?

jbkalmbach:
Ignore that last sentence. You let it handle both tuple and `len=2` in case we are not using DC2 defs, got it.

Issue #474, opened 2018-07-09
sprinkler won't produce unique galaxy_ids in DC2


danielsf:
For Twinkles, we used this line of code to produce unique identifiers for the sprinkled objects

https://github.com/LSSTDESC/Twinkles/blob/master/python/desc/twinkles/sprinkler.py#L338

This line assumes that all `galtileid`s have the same number of significant bits.  That will not necessarily be true in DC2 where `galtileid` is replaced by `galaxy_id`, and `galaxy_id==1` is a legal value.  This could cause collisions between sprinkled `galaxy_id`s and unsprinkled `galaxy_id`s.  We need to make sure that does not happen.

danielsf:
Also, we should think about whether this line

https://github.com/LSSTDESC/Twinkles/blob/master/python/desc/twinkles/sprinkler.py#L186

is going to produce useful seeds, now that `galtileid==galaxy_id` can be less than `10**8`

jbkalmbach:
After discussion with @danielsf, I will replace the `TWINKLES_ID` line that gets the image number added to it by a unique number for each Twinkles system + image combination to bring the added numbers down from 10000 to ~5000. We should have enough buffer with 5000*1024*1e11 to stay under the 9e18 integer limit.

jbkalmbach:
Since this will require some editing of the sprinkler code I want to make sure the validation code gets merged before changing anything. After that I will update this.

jbkalmbach:
This is no longer waiting on validation code but it is waiting on final DC2 galaxies.

jbkalmbach:
Closed in #480

PR #480, opened 2018-09-28
Adjust for CosmoDC2_v1.0 galaxies.

jbkalmbach:
The biggest `galaxy_id` in `cosmoDC2_v1.0_image` is ~ 1.44e10. Here I change the id scheme in the sprinkler to (`galaxy_id` + 1.5e10) * 10000 + `twinkles_system`*4 + `image_number` — where `twinkles_system`*4 + `image_number` is less than 10000 —  and when we left_shift we remain under the 2**63 limit still.

### DIFF
--- a/python/desc/twinkles/sprinkler.py
+++ b/python/desc/twinkles/sprinkler.py
@@ -266,12 +266,12 @@ class sprinkler():
                         #3 numbers to get twinklesID in the twinkles lens catalog and the remainder is
                         #the image number minus 1.
                         if not isinstance(self.defs_dict['galtileid'], tuple):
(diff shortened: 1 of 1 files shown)


jbkalmbach:
Connected to issue #474

danielsf:
Does the setting `np.random.seed` at line 209 of `sprinkler.py` need to be changed to account for the fact that `galaxy_id` has a different range than fatboy's `galtileid`?

jbkalmbach:
We could change that but since we are using cached ids anyway it should never actually run that line in DC2.

jbkalmbach:
If nobody has any issues I will merge this after 1pm Pacific today.

PR #483, opened 2018-11-05
U/danielsf/instance catalog production/181031

danielsf:
This PR updates `sprinkler.py` to reflect the new column names in the Twinkles data files.

It also increases the speed of the sprinkler by
1) only calling `find_lens_candidates` if `self.cached_sprinkling == False`
2) appending rows to `updated_catalog` in a single batched call to `np.append`, rather than multiple single-row calls.


danielsf:
No.  We have not used the `ic_validation` code.  Can you point me to an example of how it should be run?

jbkalmbach:
You should be able to use the script in the `utils` folder. Running from the command line should work as:

`python run_validation.py /path/to/catalos obsHistId /path/to/SN/SEDs`

danielsf:
@jbkalmbach Thanks for the nudge.  I had introduced a bug such that the sprinkler wasn't adding the lens galaxy for a supernova unless the supernova itself was visible.  Your script caught it.  Good job.

Happy Thanksgiving!

danielsf:
@jbkalmbach This is the branch I was talking about, in case you forgot

danielsf:
This is being supplanted by

https://github.com/LSSTDESC/Twinkles/pull/495

PR #484, opened 2018-11-29
Fixed bug that mixed lensed SNe systems with AGN hosting galaxies.

jbkalmbach:


jbkalmbach:
Also added some preliminary v1.1 catalogs but they are not to be used for final production with cosmoDC2 v1.1. Just to be used to double check that `gal_ids` do not clash with AGN.

PR #489, opened 2018-12-18
Updated caches and sprinkler inputs for CosmoDC2_v1.1.4

jbkalmbach:
* Also cleaned up old versions of cache files in data folder
* Started work on writing scripts with the notebook code to make it easier when updating sprinkler inputs and caches. Will turn it into a much easier pipeline when finished but for now is tagged as work in progress but still wanted to add so that at least the first steps are easily reproducible in the near future.


utils/catalog_production/add_glsn_properties.py
```diff
@@ -0,0 +1,286 @@
...
+
+    sed_name, magNorm, av, rv = sed_from_galacticus_mags(mag_30_list, redshift_list,
+                                                         H0, Om0, sed_min_wave, sed_wave_width, lsst_mags)
+
+    return sed_name, magNorm, av, rv
+
+
+if __name__ == "__main__":
+
+    catalog_version = sys.argv[1]
```
danielsf: Please use `argparse.ArgumentParser()` and include a `help` entry for each argument so that someone who is not us can figure out how to run these scripts.

utils/catalog_production/add_glsn_properties.py
```diff
@@ -0,0 +1,286 @@
...
+    np.savetxt('magNorm_list_sn.txt', magNorm_array)
+    np.savetxt('av_list_sn.txt', av_array)
+    np.savetxt('rv_list_sn.txt', rv_array)
+
+    test_bandpassDict = BandpassDict.loadTotalBandpassesFromFiles()
+    imsimband = Bandpass()
+    imsimband.imsimBandpass()
+
+    mag_norm_glsne = []
+    for i, idx in list(enumerate(keep_rows)):
```
danielsf: I don't think `keep_rows` is defined in this scope.  It is internal to `match_up_glsn()`.

utils/catalog_production/add_glsn_properties.py
```diff
@@ -0,0 +1,286 @@
...
+    print(len(keep_rows), len(gcr_glsn_match[:, 0]))
+    for keep_idx in range(len(keep_rows)):
+        results_dict[str(dc2_df_system['sysno'].iloc[keep_rows[keep_idx]])] = {'z':gcr_glsn_match[:, 0][keep_idx],
+                                                                               'sed_name':sed_name_array[keep_idx],
+                                                                               'magNorm':mag_norm_glsne[keep_idx]}
+
+    keep_systems = dc2_df_system['sysno'].iloc[keep_rows].values
+
+    final_df_z = []
+    final_df_lens_sed = []
```
danielsf: The `Av` and `Rv` values returned by `sed_from_galacticus_mags` is also an important part of the color fitting.  Should we be storing those in the Twinkles caches as well?

utils/catalog_production/add_om10_properties.py
```diff
@@ -0,0 +1,383 @@
...
+import pandas as pd
+from GCR import GCRQuery
+sys.path.append('/global/homes/b/brycek/DC2/sims_GCRCatSimInterface/workspace/sed_cache/')
+from SedFitter import sed_from_galacticus_mags
+from lsst.sims.photUtils import Sed, Bandpass, BandpassDict
+
+
+def get_sl2s_data():
+    filename = '../../data/SonnenfeldEtal2013_Table3.csv'
+    ! wc -l $filename
```
danielsf: I do not believe this will run in a python script.

utils/catalog_production/add_om10_properties.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+def estimate_stellar_masses_om10():
+
+    # Instantiate an empiriciSN worker object:
+    empiricist = empiriciSN.Empiricist()
+    X, Xerr = get_sl2s_data()
+
+    # Load in cached om10 catalog
+    hdulist = fits.open('../../data/om10_qso_mock.fits')
```
danielsf: This path is no longer correct (because we are only one directory removed from `$TWINKLES_DIR` in this script).

It is probably safer to use
```
os.path.join(os.environ['TWINKLES_DIR'], 'data', 'om10_qso_mock.fits')
```
which is not dependent on where you are when you run this script.

utils/catalog_production/add_om10_properties.py
```diff
@@ -0,0 +1,383 @@
...
+    twinkles_lenses = hdulist[1].data
+
+    # Predict a mass for each galaxy:
+    np.random.seed(0)
```
danielsf: Again: do you mind using `np.random.RandomState`?

utils/catalog_production/add_om10_properties.py
```diff
@@ -0,0 +1,383 @@
...
+        twinkles_log_m_1comp = np.append(twinkles_log_m_1comp,log_m)
+    
+    return twinkles_lenses, log_m, twinkles_log_m_1comp
+
+
+def get_catalog(catalog, twinkles_lenses, twinkles_log_m_1comp):
+
+    gcr_om10_match = []
+    err = 0
+    np.random.seed(10)
```
danielsf: `np.random.RandomState`, please

utils/catalog_production/add_om10_properties.py
danielsf: Actually: it doesn't look like this random number generator gets used...

utils/catalog_production/add_glsn_properties.py
```diff
@@ -0,0 +1,286 @@
...
+    data_df = pd.DataFrame(data)
+    data_df.to_csv('sn_matching_checkpoint_1.csv', index=False)
+
+
+def match_up_glsn(dc2_df_system, data_df):
+
+
+    gcr_glsn_match = []
+    err = 0
+    np.random.seed(10)
```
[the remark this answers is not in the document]
danielsf: Alternatively: I don't know if you want to make an `rng` kwarg in all of these methods so that you can pass a single random number generator around.  That would make sure that every random number draw is random wrt previous draws in other methods.  This comment would apply to the methods in the other scripts, as well.

utils/catalog_production/add_om10_properties.py
danielsf: this random number generator also does not appear to be in use

utils/catalog_production/add_om10_properties.py
```diff
@@ -0,0 +1,383 @@
...
+    redshift_list = np.array(redshift_list)
+
+    sed_name, magNorm, av, rv = sed_from_galacticus_mags(mag_30_list, redshift_list,
+                                                         H0, Om0, sed_min_wave, sed_wave_width, lsst_mags)
+
+    return np.array(sed_name), np.array(magNorm), np.array(av), np.array(rv)
+
+if __name__ == "__main__":
+
+    sys.argv[1] = catalog_version
```
danielsf: Again, we should probably use `argparse` here

jbkalmbach:
Good point. I did notice those values. I will include them in the catalogs and we will include that in the `sprinkler.py` update.

danielsf:
Okay.  Let me know when you have merged this and I will rebase my branch.

jbkalmbach:
Caches now include av, rv values. Thanks for looking at the utils scripts, I made some changes and made notes of others for later.

PR #495, opened 2019-01-04
U/danielsf/instance catalog production/181221

danielsf:
I think this is ready for production.  I think you looked at most of the code already, but so much hacking has gone on, I think it would be best if you gave it another look over before we merged it.


python/desc/twinkles/sprinkler.py
```diff
@@ -185,255 +184,317 @@ def __init__(self, catsim_cat, visit_mjd, specFileMap, sed_path,
...
+            # Append the lens galaxy
+            # For each image, append the lens images
+            default_lensrow = None
+            if newlens['NIMG'] > 0:
+                default_lensrow = row.copy()
+                default_lensrow[self.defs_dict['galaxyDisk_majorAxis']] = 0.0
+                default_lensrow[self.defs_dict['galaxyDisk_minorAxis']] = 0.0
+                default_lensrow[self.defs_dict['galaxyDisk_positionAngle']] = 0.0
+                default_lensrow[self.defs_dict['galaxyDisk_internalAv']] = 0.0
+                default_lensrow[self.defs_dict['galaxyDisk_magNorm']] = 999. #np.nan To be fixed post run1.1
```
jbkalmbach: Should we remove these notes to ourselves about handling the magNorms? These are working so that in the final instance catalogs we only have the disk and bulge entries from `bulge/disk_agn_host` right?

python/desc/twinkles/sprinkler.py
```diff
@@ -185,255 +184,317 @@ def __init__(self, catsim_cat, visit_mjd, specFileMap, sed_path,
...
+                    lensrow[self.defs_dict['galaxyAgn_is_sprinkled']] = 1
+                    lensrow[self.defs_dict['galaxyBulge_is_sprinkled']] = 1
+                    lensrow[self.defs_dict['galaxyDisk_is_sprinkled']] = 1
+
+                #To get back twinklesID in lens catalog from phosim catalog id number
+                #just use np.right_shift(phosimID-28, 10). Take the floor of the last
+                #3 numbers to get twinklesID in the twinkles lens catalog and the remainder is
+                #the image number minus 1.
+                if not isinstance(self.defs_dict['galtileid'], tuple):
+                    lensrow[self.defs_dict['galtileid']] = ((lensrow[self.defs_dict['galtileid']]+int(1.5e10))*100000 +
```
jbkalmbach: This is different than the number scheme I tested because I didn't take into account that we would have > 4 images in some systems. I think this takes us potentially over 2**63. Is this a problem?

python/desc/twinkles/sprinkler.py
```diff
@@ -185,255 +184,317 @@ def __init__(self, catsim_cat, visit_mjd, specFileMap, sed_path,
...
-                    row[self.defs_dict['galaxyBulge_positionAngle']] = newlens['PHIE']*(-1.0)*np.pi/180.0
-
-                    if self.logging_is_sprinkled:
-                        row[self.defs_dict['galaxyAgn_is_sprinkled']] = 1
-                        row[self.defs_dict['galaxyBulge_is_sprinkled']] = 1
-                        row[self.defs_dict['galaxyDisk_is_sprinkled']] = 1
-
-                    #Replace original entry with new entry
-                    updated_catalog[rowNum] = row
+                rng = np.random.RandomState(galtileid % (2^32 -1))
```
jbkalmbach: Should we change these 2^32 values with 2^64 values due to the integer limit change?

python/desc/twinkles/sprinkler.py
```diff
@@ -185,255 +184,317 @@ def __init__(self, catsim_cat, visit_mjd, specFileMap, sed_path,
...
+            default_lensrow[self.defs_dict['galaxyBulge_magNorm']] = 999. #np.nan To be fixed post run1.1
+            default_lensrow[self.defs_dict['galaxyBulge_sedFilename']] = None
+            varString = 'None'
+            default_lensrow[self.defs_dict['galaxyAgn_varParamStr']] = varString
+
+            for i in range(len(use_df)):
+                lensrow = default_lensrow.copy()
+                delta_ra = np.radians(use_df['x'].iloc[i] / 3600.0)
+                delta_dec = np.radians(use_df['y'].iloc[i] / 3600.0)
+                if use_system == 1953:
```
jbkalmbach: Is this meant to print in production? It was probably for testing/debugging I'm guessing.

danielsf: Good catch.  I will remove that

python/desc/twinkles/sprinkler.py
danielsf: What are we trying to achieve here?  Is the idea that each AGN should have its own unique random seed?

`np.random.RandomState` requires the seed to be a 32 bit integer.  `2**32=4.3*10**9`, which is less than the maximum `galaxy_id`, so we do need to cast this into a 32 bit integer.  That being said, `^` is a logical operator.  Should this line be `rng=np.random.RandomState(galtileid % (2**32-1))`?

python/desc/twinkles/sprinkler.py
danielsf: So: the concern is that we want to make sure that the galaxies that are in the catalog pre-sprinkler are totally removed...

I will add a test to the InstanceCatalog validation script to just verify that none of the values of `galaxy_id` specified in the `cosmoDC2_v1.1.4_agn_cache.csv` appear in the InstanceCatalogs.  Will that address the concern?

jbkalmbach: Yeah, I think that would be good. Thanks.

python/desc/twinkles/sprinkler.py
jbkalmbach: We also need to include the bit shift correct? But I think we are still ok since it looks like that would get us to `~3*10**18`.

python/desc/twinkles/sprinkler.py
jbkalmbach: Now that I think about it yeah, we just wanted unique seeds but also a way to back everything out if we needed to. So, I think yes, just  `rng=np.random.RandomState(galtileid % (2**32-1))` is what we want.

python/desc/twinkles/sprinkler.py
danielsf: good save

but: yes, I think it is still safe as written

python/desc/twinkles/sprinkler.py
danielsf: I pushed a change to `validate_ic.py` yesterday.  Verifying that none of the replaced `galaxy_id`s appear isn't quite right, because the lens galaxies keep the old `galaxy_id`.  It's only the multiply-imaged AGN that get new `galaxy_id`s.  So, the new test

- Verifies that no `galaxy_id`s greater than 10**11 exist in the Sersic profile InstanceCatalogs (since this would indicate that the bulge/disk components of the multiply lensed AGN were making it into the InstanceCatalog)

-Verifies that none of the point sources have `magNorm>900` or `magNorm==None`, since that would indicate that CatSim was not correctly handling the placeholder magNorms.

The commit in question is here

https://github.com/LSSTDESC/Twinkles/pull/495/commits/2d6f5c6f4cec43fdb3d0b836d17587826b2a7e54

Do you agree with my reasoning?

jbkalmbach: Yes, this looks right. The host bulge/disk components from Matt's code go in a separate file, right?

danielsf: Yes.  There is a separate InstanceCatalog file for the FITS-based components

danielsf:
I am going to merge this at 5pm PST today, unless anyone objects

PR #496, opened 2019-05-06
U/danielsf/validation

danielsf:
This is not a high priority, but I reshuffled the validation code so that we can import the twinkles validation tools into the general InstanceCatalog validation scripts I have in sims_GCRCatSimInterface


jbkalmbach (approved):
Looks good.

danielsf:
@jbkalmbach Do you mind taking one more look.

I was getting failures in the validation of sprinkled SNe magnitudes.  I decided our test was too strict.  I altered the test so that it verifies that the discrepancy between what magNorm is and what magNorm should be results in a flux discrepancy no more than 0.1*flux_noise for the object.

danielsf:
I do realize now that I may have conflated the imSimBandpass with the actual LSST bandpasses when calculating signal to noise.  This might be something we want to revisit in the future.

jbkalmbach:
Looks good. Merging to master.

